PR #7, opened 2022-07-28
Changed column names & added example to dbpedia

rohanoai:
Can now call `openai tools fine_tunes.prepare_data -f dbpedia_samples.jsonl` without bugs.


shyamal-anadkat:
@rohanoai do you still want to proceed with this change?

ted-at-openai:
> @rohanoai do you still want to proceed with this change?

My hesitation in merging this change is that it turns descriptive column names (text, category) into abstract fine-tuning-specific column names (prompt, completion). Given that our current example doesn't use fine-tuning, it feels a little odd to be importing a data table that names categories as completions.

My preference would be to keep the adjustment to the data class representation (so that it's less likely to cause problems with fine-tuning should someone wish to) but not to change the column names to prompt & completion.

Thoughts?

PR #8, opened 2022-08-03
organizes links in README

shyamal-anadkat:


ted-at-openai (approved):
Thanks for the contribution!

PR #33, opened 2022-11-05
updates DALL-E API example

ted-at-openai:
Updates the previous DALL-E example in a number of ways:
- Renames the example file
- Updates the parameters to match the launched API
- Uses the OpenAI Python library interface
- Other small changes like linting, updated comments, and updated text


ted-at-openai:
FYI @colin-jarvis: I updated the DALL-E example to follow the updated API

rememberlenny:
Thank you!

PR #35, opened 2022-11-15
adds unit test example

ted-at-openai:
Adds a unit test example that illustrates how to use multiple prompts and some other flourishes to get more reliable performance from GPT-3.


kevinjyh:
The ipynb you wrote is really great for me, it helps me understand pytest and related testing concepts. However, when I use the tested function "def is_palindrome(s)" from the tutorial, there will be non-green output text (newline=\n) at the end, but no error message, so I can't figure out what the problem is. If I use my own tested function, I often get the error message "ChunkedEncodingError: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))" which occurs in the "{GENERATED IN STEP 3}" stage. Do you have any resources that can help me solve my confusion? Because I really love this ipynb, it is very helpful to me.

PR #36, opened 2022-11-18
Added .default scope to URI to get token using DefaultAzureCredential

achandmsft:
Fixing issue encountered when using this in some contexts (like virtual environments in notebooks). @sorinsuciu-msft FYI


ted-at-openai (approved):
Apologies for the delay on my review. (Was on vacation and it slipped by me.) Merging today. :)

PR #46, opened 2022-12-23
Only get_embedding once

scottleibrand:
https://github.com/openai/openai-cookbook/commit/fd181ec78fa57e14583e27915ccdad3a084296ae updated what was previously two different get_embedding() calls w/ different models to be two identical calls. Rather than doing the same API call twice in a row, we can now just set the second variable to be equal to the first one.


shyamal-anadkat (approved):

shyamal-anadkat:
thanks!

PR #47, opened 2022-12-23
line-wrap comment in code block for readability

scottleibrand:
The github UI doesn't allow you to horizontally scroll code blocks, and this one is too long to display even with a maximized window.

Let's line-wrap the comment for readability.


scottleibrand:
This change is also included in #48, so if that can be merged as-is, this one should be closed to avoid any merge conflicts.

aniknandi (approved):

ted-at-openai:
Moved the data from CDN to repo, so URL is no longer needed. Will close.

PR #48, opened 2022-12-23
Workarounds for CSV on CDN still saying babbage_search

scottleibrand:
If you download the .csv from the provided CDN URL, it was generated before the model consolidation and still has babbage_search in its header row.

Commented two different workarounds for that. This commit can be reverted once https://cdn.openai.com/API/examples/data/fine_food_reviews_with_embeddings_1k.csv is re-uploaded using the updated code at https://github.com/openai/openai-cookbook/blob/main/examples/Obtain_dataset.ipynb


ted-at-openai:
Hey Scott! Long time since DAGGRE. :)

Thanks for the PR and apologies for this error living stale over the winter break. I've fixed it in #66. Precomputed embeddings are now ada-002, and stored in the data folder of the repo rather than on our CDN.

Will close this PR. Let me know if any other issues remain.

scottleibrand:
Lol, nice: small world.

Thanks for fixing it properly. This is exactly the way I like to see my PRs closed instead of merged. ;-)

PR #49, opened 2022-12-27
Updated Obtain_dataset.ipynb to reduce the number of API calls

damienbose:
Also, 'ada_similarity' and 'ada_search' should be replaced by the column name 'embedding' to reduce confusion.


ted-at-openai:
Thank you! Fixed in https://github.com/openai/openai-cookbook/pull/66.

PR #62, opened 2023-01-10
added some points and remove some extra spaces.

sumanpaikdev:


ted-at-openai:
Thanks for the contribution, but I will decline to merge for now. I prefer the paragraphs to lists (although I acknowledge they take more space), and I don't want to add newline characters into a quotation that does not have them (although I acknowledge it's harder to read without wrapping on GitHub.

PR #77, opened 2023-01-16
Replaced non-existent model with the new version.

Pankaj-Baranwal:
davinci-instruct-beta-v2 model doesn't exist anymore. It has been replaced by davinci-instruct-beta-v3


ted-at-openai:
Thanks for the change. I'm happy to merge as-is, but my two concerns are:
- the most up-to-date instruction-following model is now `text-davinci-003`
- after this update, the outputs in the notebook no longer correspond to the model written in the code (as it doesn't look like the output was regenerated)

Ultimate options:
1. Merge as-is
2. Add a note that the outputs are from an older model
3. Replace with text-davinci-003 and re-run the notebook (if it's even possible) to get the actual outputs of the model

#1 is easiest, so I'll go with that for now.

PR #81, opened 2023-01-18
Add Qdrant as another example of vector database

kacperlukawski:
I added Qdrant vector database to the notebook using the same approach as you did for Pinecone and Weaviate. Due to some rate limits and other issues, I could not create the embeddings for all the documents so it may require relaunching. However, all the required logic is already in place, with a corresponding `docker-compose.yaml` file to run Qdrant in a container.


kacperlukawski:
@colin-openai I'd appreciate your review on that.

colin-openai (approved):
Reviewed, nice and concise and ran locally. Approved

PR #84, opened 2023-01-19
Add example for handling long input texts

filipeabperes:


filipeabperes:
Should we renorm the embedding vector after doing the weighted average? 🤔

ted-at-openai:
>Should we renorm the embedding vector after doing the weighted average? 🤔

Good idea - I'll do that now.

I also committed some changes above to polish the text. I also re-ran the notebook to clean up the cell numbers (1, 2, 3, ...) as well as align the outputs with the code (I think the print statements had been reworded, but not re-executed).

If my suggested changes look good to you, happy to merge.

ted-at-openai (approved):

filipeabperes:
Thanks! Everything looks good to me, merged.

PR #92, opened 2023-01-29
Encode separator for completions endpoint

Androbin:
`SEPARATOR` is being used to construct a prompt for use with the completions endpoint


ted-at-openai:
Good catch!

PR #95, opened 2023-01-30
Updated a markdown in Clustering.ipynb

akshayvkt:
Updated the penultimate markdown to align with the code in the following cell


ted-at-openai:
Good catch - thank you!

PR #104, opened 2023-02-03
Fix UnicodeEncodeError

khaled-wsa:
This pull request fixes the UnicodeEncodeError in the code by adding encoding="utf-8" to the open() function calls.

### DIFF
--- a/solutions/web_crawl_Q&A/web-qa.py
+++ b/solutions/web_crawl_Q&A/web-qa.py
@@ -135,7 +135,7 @@ def crawl(url):
         print(url) # for debugging and to see the progress
 
         # Save text from the url to a <url>.txt file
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Hey! Thanks for the quick PR, can you share how this error came up? I ran this a few times locally with no issues.

khaled-wsa:
> 
`File "C:\Users\---\AppData\Local\Programs\Python\Python310\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 172-176: character maps to <undefined>`

I think this error may have been caused by the website, since it occurs when trying to encode characters that are not supported by the current encoding. This may have been due to processing Hebrew or Arabic characters. To solve this issue, I added encoding="utf-8" to the open() function calls, which should support those characters and it resolved the error.

PR #109, opened 2023-02-04
Remove unmatched closing brackets ")"

arinbjornk:
Remove unmatched closing brackets ")". 

The added unmatched ")" causes a syntax error. Removing it fixes the error.


logankilpatrick:
Good catch, thank you!

PR #114, opened 2023-02-06
Fix weighted average

filipeabperes:


filipeabperes:
@ted-at-openai good to merge?

PR #122, opened 2023-02-09
Merging Weaviate notebooks to main

colin-openai:
Worth one additional review to ensure main notebook runs end-to-end


ted-at-openai:
Suggestion: in the notebook, instead of saying the file is "pretty big" say it's "~700 MB". I expect a term like "pretty big" means different things to different people, and it will be hard for them to know what to expect.

sebawita:
Hey @ted-at-openai, thank you for the feedback.

I am OK with "pretty big" => "~700 MB" change.

Is this the only change you need from us?

ted-at-openai:
Still running through the example notebook. Will drop comments as I have them. :)

ted-at-openai:
> Hey @ted-at-openai, thank you for the feedback.
> 
> I am OK with "pretty big" => "~700 MB" change.
> 
> Is this the only change you need from us?

Still running through the example notebook. Will drop comments as I have them. :)

ted-at-openai:
"The other vector database option" -> "Another vector database option" (I assume this was written when there were just two)

ted-at-openai:
I tried running `docker-compose up -d` and immediately hit a blocker: `no configuration file provided: not found`

Probably something simple, but if I ran into it, others will too.

ted-at-openai:
I think instead of "You can start Weaviate up locally by navigating to this directory" it should say the actual directory. Phrases like "this directory" can always be a bit ambiguous. Here I think it means the directory of the yml file, not the directory of the notebook I was reading (my original interpretation).

ted-at-openai:
Next error I hit was `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` Do we need clearer instructions here?

ted-at-openai:
> Next error I hit was `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` Do we need clearer instructions here?

Solution to this error was to start Docker first. Maybe rephrase previous instruction to "After starting Docker, ..."

ted-at-openai:
(sorry for all the comments here instead of in-line. looks like I can only do in-line comments for the lines affected by the latest commits and not the earlier ones)

ted-at-openai:
Otherwise, everything ran perfectly. Code and output look clean. This is fantastic!

ted-at-openai (approved):

sebawita:
Hi @ted-at-openai 
Thank you for your feedback, your changes (you were 100% spot on) and merging the PR.
This looks amazing.

Any chance we could update the Docker link to - https://www.docker.com (instead of https://www.docker.com/pricing)?

ted-at-openai:
Updated: https://github.com/openai/openai-cookbook/pull/129

PR #132, opened 2023-02-14
Update README.md to add recently added section with links

logankilpatrick:


ted-at-openai (approved):
Looks good, feel free to merge.

Possible ideas for improvement:

- Order items from most recent to least recent (not sure what current ordering is)
- Add [new]/sparkles to the same items in the overall table of contents (not actually sure if this would look good or not)
- The text feels a little bit wordy - could be tightened up or even possibly deleted (hopefully 'recently added' is self-explanatory)

But none of these potential improvements are blockers. Thanks for adding this!

logankilpatrick:
Fixed ordering and removed the extra wording! I also think adding the emojis to the list would be  very helpful

PR #155, opened 2023-02-23
Catch the exception and continue processing the next link

ashokmanghat:
As of the now, the web crawler stops processing the remaining items from the link queue if any of the link is invalid. To fix the issue a new exception handling code was added to catch the exception and procced with the next item in the link queue.


ted-at-openai:
@isafulf Do you want to address/review this?

arifhossain512 (approved):

simonpfish (approved):
Makes sense! Thank you for the contribution, this makes the code more robust. 

@logankilpatrick could be worth considering updating the code on the tutorial site to show this as well

PR #156, opened 2023-02-23
Changed "the the endpoint" to "to the endpoint"

JJnotJimmyJohn:
Line 14 -> I think it should be "call made to the endpoint" instead of "call made the the endpoint"


logankilpatrick (approved):

logankilpatrick:
Thank you!

PR #184, opened 2023-03-05
Handle a null answer from the server.

sammyjava:
I'm finding that the server sometimes returns a null answer. This crashes the client app. This PR adds logic to spit out the response "I did not receive an answer!" when that happens. It isn't clear to me why the server returns a null answer rather than "I couldn't find the answer to that question in your files." It seems to vary with server restarts.

### DIFF
--- a/apps/file-q-and-a/nextjs-with-flask-server/client/src/components/FileQandAArea.tsx
+++ b/apps/file-q-and-a/nextjs-with-flask-server/client/src/components/FileQandAArea.tsx
@@ -108,14 +108,11 @@ function FileQandAArea(props: FileQandAAreaProps) {
           className="mb-8"
         >
           {/* answer from files */}
-          {answer && (
+          {answer ? (
             <div className="">
               <ReactMarkdown className="prose" linkTarget="_blank">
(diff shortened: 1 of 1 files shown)


soheil (commented):
Thanks for submitting this PR. The change is straight-forward and will definitely fix the client crashes. However, there is one thing I'd like to point out - Instead of spitting out "I did not receive an answer!" for a null response, can we add a more meaningful error message like "No matching result found" or "Unable to find the answer"? This will help the user better understand what has happened and will help them know how to move forward. I'd suggest we update the message accordingly. Other than that, the code looks good to me.

sammyjava:
Good point. Actually, it should probably say that it received nothing, like "No response was returned." Agree? I'll update if so.

simonpfish:

Unfortunately, due to current workload constraints and priorities, our team doesn't have the bandwidth to thoroughly review and integrate new pull requests for these full-stack examples. Please understand that this isn't a reflection on the quality of your work, which we genuinely appreciate. It's simply a matter of resource allocation and current focus.

Thank you again for your understanding and for your contribution.

PR #189, opened 2023-03-06
Update techniques_to_improve_reliability.md

goodship1:
Small grammar changes


ted-at-openai:
Full sentences need to be capitalized but fragments do not. I'll leave as is. Thanks for the PR, though!

PR #202, opened 2023-03-09
Weaviate: update links, minor corrections

dandv:
The suffix at the end of the cluster name is a new addition that the WCS now mandates when creating sandboxes.


colin-openai (approved):
Reviewed, mostly minor text changes and the area where code was reformatted still runs. All good.

Issue #79, opened 2023-01-16
Document Library Pre-Processing


CoreyRN:
Hello all,

Would it be at all possible to provide an example of document pre-processing where the dataset is not being imported from Wikipedia, instead through an individualized standard csv file?

For no less than 72 hours now over the past week I've been trying to complete the question and answering tutorial (https://github.com/openai/openai-cookbook/blob/main/examples/Question_answering_using_embeddings.ipynb) using my own dataset. The issue is even though I downloaded the example csv file, and copied my own data into the csv file and re-saved it, I cannot get the dataset to run using the code. The code runs perfectly fine if I use the sample data set but when I try to run the sample data set with my data replacing it (even ensuring it is always saved as a csv file), it always errors past line 48. I have tried changing the data types in the columns using python, I've tried removing any special characters, I've tried, I kid you not, about three days worth of fixes with no luck. ChatGPT is now repeating recommendations without any success unfortunately. 

I continually receive this error:

`ValueError                                Traceback (most recent call last)
Cell In [74], line 1
----> 1 prompt = construct_prompt(
      2     "What is a WOC Nurse?",
      3     document_embeddings,
      4     df
      5 )
      7 print("===\n", prompt)

Cell In [73], line 16, in construct_prompt(question, context_embeddings, df)
     13 document_section = df.loc[section_index]
     15 chosen_sections_len += document_section.tokens + separator_len
---> 16 if chosen_sections_len > MAX_SECTION_LEN:
     17     break
     19 chosen_sections.append(SEPARATOR + document_section.content.replace("\n", " "))

File /shared-libs/python3.9/py/lib/python3.9/site-packages/pandas/core/generic.py:1442, in NDFrame.__nonzero__(self)
   1440 @final
   1441 def __nonzero__(self):
-> 1442     raise ValueError(
   1443         f"The truth value of a {type(self).__name__} is ambiguous. "
   1444         "Use a.empty, a.bool(), a.item(), a.any() or a.all()."
   1445     )

ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
`

You can see in the dataset found here: https://docs.google.com/spreadsheets/d/e/2PACX-1vSs9Ok5FUrhAOu_BnpLwV63bwpLylRtUWBDE7onAX1zrZW0Sz4gBEtBN-KtsBiC1DhKyhhZjNXfNf0i/pub?output=csv

[...]

PR #208, opened 2023-03-10
fix iteration error

justanotherlad:
array iteration error: sentences up till max_len of token exceeds should be returned

Here's a screenshot of before and after the change (vertically in terminal) in the code. Clearly, the below should be returned: (The numbers vertically represent _*current tokens*_ and _*total_tokens*_ number for each iteration.

Screenshot from 2023-03-07 21-16-59


ted-at-openai:
Great catch and thanks for the super helpful documentation.

Would it be easier just to write `sentences[:i]` instead of `sentences[:i+1][:-1]`? Seems odd that the list is being sliced twice in the first place.

But happy to merge as is if you don't feel like doing any extra work. Your work so far is already appreciated.

justanotherlad:
Also, please see my comment on [ Document Library Pre-Processing #79 ](https://github.com/openai/openai-cookbook/issues/79#issuecomment-1467670185). I think there might be a bug there.

justanotherlad:
@ted-at-openai Do you want to close/merge this PR?

ted-at-openai:
Thanks! Merged.

PR #227, opened 2023-03-16
docs: update dalle example file

liuliuOD:
1. add a terminal command sample about how to export OPENAI_API_KEY to environment


ted-at-openai:
Thanks - that should help people with that step.

PR #228, opened 2023-03-16
fix prompt list counter

MrZilinXiao:
When counting token consumption for multiple prompts, completion tokens should be multiplied by the length of the prompt list.


ted-at-openai:
Great catch - thank you!

PR #240, opened 2023-03-17
Fixes logging to use Flask's default logging config

zbskii:
Flask sets up its logger and doesn't touch the root python logger. This patch centralizes the logging code in app.py and copies the default flask logger config to the root logger


simonpfish:

Unfortunately, due to current workload constraints and priorities, our team doesn't have the bandwidth to thoroughly review and integrate new pull requests for these full-stack examples.

Thank you again for your understanding and for your contribution.

PR #243, opened 2023-03-18
num_tokens_from_messages off by 1 token count

jwchang0206:
Today I found that the calculation was wrong when I was calculating the token size for `gpt-3.5-turbo`

### DIFF
--- a/examples/How_to_format_inputs_to_ChatGPT_models.ipynb
+++ b/examples/How_to_format_inputs_to_ChatGPT_models.ipynb
@@ -459,7 +459,7 @@
     "                num_tokens += len(encoding.encode(value))\n",
     "                if key == \"name\":  # if there's a name, the role is omitted\n",
     "                    num_tokens += -1  # role is always required and always 1 token\n",
-    "        num_tokens += 2  # every reply is primed with <im_start>assistant\n",
+    "        num_tokens += 3  # every reply is primed with <im_start>assistant\n",
     "        return num_tokens\n",
     "    else:\n",
     "        raise NotImplementedError(f\"\"\"num_tokens_from_messages() is not presently implemented for model {model}.\n",


jwchang0206:
@ted-at-openai

ted-at-openai:
Can you give an example? The counts match in the example in the notebook.

jwchang0206:
```typescript
const numberOfTokensFromTurboMessages = (
  messages: Array<ChatCompletionRequestMessage>,
) => {
  // Extend existing encoding with custom special tokens
  const encoding = encoding_for_model('gpt-3.5-turbo');
  let numTokens = 0;
  messages.forEach((message) => {
    numTokens += 4; // every message follows <im_start>{role/name}\n{content}<im_end>\n
    Object.keys(message).forEach((key) => {
      const value = message[
        key as keyof ChatCompletionRequestMessage
      ] as string;
      numTokens += encoding.encode(value).length;
      if (key === 'name') {
        // if there's a name, the role is omitted
        numTokens += -1; // role is always required and always 1 token
      }
    });
  });
  numTokens += 3; // every reply is primed with <im_start>assistant
  encoding.free();
  return numTokens;
};

const tokensLength = numberOfTokensFromTurboMessages(messages);
const modelMaxTokens = 4096 - tokensLength;
await openai.createChatCompletion({
      model: 'gpt-3.5-turbo',
      messages, // messages: Array<ChatCompletionRequestMessage>,
      max_tokens: maxTokens,
     ...
});
```

@ted-at-openai 
I simplified the code that I'm using and it used to give off by 1 error and now it works with the change.

jwchang0206:
https://github.com/openai/openai-cookbook/pull/254

ted-at-openai:
FYI, we just changed how we count tokens on our backend. Total token consumption is the same, but we're now counting one more token as part of the prompt and one less for the completion. I'll update the code. Be aware that this may change again in the future.

ted-at-openai:
Fixed in #278

PR #244, opened 2023-03-18
Add chat endpoint support

herobs:
1. Use regex to match the API endpoint component.
2. Add support for the chat/completions endpoint.
3. Replace eval with `json.loads` for more robustness and safety.


alexsio27444 (approved):

ted-at-openai:
Excellent PR. Appreciate adding chat/completions, and also for using json.loads instead of eval. Happy to merge.

(One follow-up change I'll make is to update the chat token counting function, as the token counts change depending on gpt-3.5-turbo vs gpt-4. Should probably live in the OpenAI python library as a utility function, but calling that here is less educational for someone trying to write equivalent functionality in another language.)

ted-at-openai (approved):

PR #246, opened 2023-03-18
Fix typo in `techniques_to_improve_reliability.md`

OTheDev:
Propose the following change: 
 - "For example, if you were you can try more..." -> "For example, you can try more..."

I think that "wanted" instead of "were" was originally intended. However, I think "if you wanted" is unnecessary here.


ted-at-openai (approved):

ted-at-openai:
Great catch - thank you!

PR #254, opened 2023-03-20
updates token counting function to work with gpt-4

ted-at-openai:


ted-at-openai:
> Hi @ted-at-openai.
> 
> When you originally submitted this PR, I was able to confirm the token counting examples you had provided.
> 
> However, I just ran them again today, and they no longer hold. The number of prompt tokens returned from the completion is one higher than that calculated from `num_tokens_from_messages()` (for both models).
> 
> I first noticed this in some of my production code, where I make an assertion to ensure the pre-calculated number of prompt tokens is equal to that returned from the completion.
> 
> Did something change recently?
> 
> Thanks!
> 
> Owain

Thanks for catching. We just changed how we count tokens on our backend. Total token consumption is the same, but we're now counting one more token as part of the prompt and one less for the completion. I'll update the code. Be aware that this may change again in the future.

ted-at-openai:
100% understand the concern. We have zero intention of ever switching out behavior on a snapshotted model. On reflection, I think my statement was unnecessarily broad. I'll submit another PR to narrow it. In this case, nothing changed except a token that was counted as part of the completion is now counted as part of the prompt instead. In the future, the format may change, but that will very likely only happen as part of a new snapshotted model. I'm afraid to say nothing will ever ever change, as I wouldn't want to tie our hands to not make a small bug fix of one sort or another. But we have zero intention of ever switching behavior of models in production.

PR #255, opened 2023-03-20
Named entity recognition methods(NER) using gpt

monk1337:
Examples of different Named entity recognition(NER) scenarios

1) Without any training data
2) With Custom Tags/Entities
3) With One shot Example
4) With a few shot Examples
5) With some Domain Knowledge

### DIFF
--- a/examples/Named_entity_recognition_methods_using_gpt.ipynb
+++ b/examples/Named_entity_recognition_methods_using_gpt.ipynb
@@ -1,0 +1,482 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "provenance": []
+    },
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3"
+    },
+    "language_info": {
+      "name": "python"
+    },
+    "gpuClass": "standard"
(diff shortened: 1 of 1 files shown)


ted-at-openai (commented):
Since this already exists in the Promptify repo, feels a bit unnecessary to duplicate it here.

That said, I'd love to better support and display developer tools like Promptify.

What do you think about a section on the main README that links to Promptify and similar tools?

monk1337:
Thank you for your response! That's a nice idea to include a section on the main README that links to Promptify and similar tools, which would undoubtedly increase their visibility within the community.

Regarding the PR, I believe that adding various NLP examples, such as the NER demo, to the openai-cookbook repo would provide a more convenient and engaging experience for users who visit the repository for inspiration and guidance. This addition would position the repo as a go-to resource for exploring the diverse NLP tasks that GPT can tackle, potentially encouraging users to experiment and innovate with the technology.

For those seeking more in-depth information, API references, or advanced applications, we can direct them to the respective repositories to explore further.

ted-at-openai:
Thanks! I added a link to Promptify in a new README section, as part of this PR: https://github.com/openai/openai-cookbook/pull/442

PR #259, opened 2023-03-21
Zilliz integration and Milvus bugfixes

filip-halt:
Included the Zilliz vector database, Milvus bugfixes, and more notebooks. 

Also this merge aims to clear the notebook outputs of the shared notebook. Every pull is deleting other peoples outputs and including their own. I believe the best route is to just not have outputs in the notebook instead of everyone overwriting each others work.

### DIFF
--- a/examples/vector_databases/Using_vector_databases_for_embeddings_search.ipynb
+++ b/examples/vector_databases/Using_vector_databases_for_embeddings_search.ipynb
@@ -78,19 +78,10 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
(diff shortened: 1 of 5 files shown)


filip-halt:
> 

It might be worth having a standard output near the end that shows exactly what should be returned. The different sections just show how to get that input using each vector store.

filip-halt:
Is there anything that needs to be changed to get this merged? @ted-at-openai

ted-at-openai:
Merged! Thanks for your patience. I think I'm going to split up the single notebook into one for each vector DB, to help keep things manageable. Hopefully can get that done tomorrow.

PR #277, opened 2023-03-25
Add Typesense example to vector databases

jasonbosco:
👋 Following up from my email conversation with @logankilpatrick, I've added an example of how to use [Typesense](https://typesense.org) as a vector datastore and implement similarity search.


ted-at-openai:
Looks good; thanks for the contribution.

Having one notebook for all examples feels crowded to me. I think it's going to make more sense to split this single notebook for all vector databases into separate notebooks for each. I'll likely merge your PR and then submit a follow-up PR to split each provider (including Typesense) into its own file. Does that sound reasonable? Was there any particular feedback you wanted on this PR? (I haven't run it myself but will try to tomorrow if it's not too hard to set up.)

PR #286, opened 2023-03-26
Update README.md to add Codespaces deploy button

logankilpatrick:
@ted-at-openai hopefully this would help people who don't want to run all the code locally, what do you think?


ted-at-openai:
Oh cool!

When I click the link, I get a 404, though.

PR #294, opened 2023-03-28
Added sharptoken as example 

dmitry-brazhenko:
Before my change:

After my change:

I implemented and tested a lib that produces same tokenization: https://github.com/dmitry-brazhenko/SharpToken


ted-at-openai:
Looks fantastic. Happy to merge.

PR #300, opened 2023-03-30
pt-BR translation of README.md

PedroFellipeAntunes:
Saw a lack of pt-BR translated content around OpenAI and wanted to introduce this small contribution to the OpenAI Cookbook.

As stated this is simply a direct translation of the original README.md file to brazilian portuguese, keeping all the information from the original while avoiding confusing interpretations of some technical terms that do not translate properly.


logankilpatrick:
Hey! Thank you for taking the time to do this, we unfortunately cannot maintain language translations at the moment, but thank you for suggesting this!

PR #303, opened 2023-03-31
Upload file size error default 4 mb extended to 30mb

arifhossain512:
## Description
# API/process-files exceed the default file size that was set to default 4 MB by process.ts. 

## Motivation and Context
# i changes that default size to a fixed size for API to handle a size equal to what was stated in the fronted to be 30 MB exact.

### none


simonpfish:

Unfortunately, due to current workload constraints and priorities, our team doesn't have the bandwidth to thoroughly review and integrate this pull request. Thank you again for understanding and for your contribution.

PR #308, opened 2023-03-31
Modify example to make it reproducible

franfram:


arifhossain512 (approved):

ted-at-openai:
Thanks! Seems unnecessary to have folks download the dataset when a jsonl is included with the repo. What do you think about including the code for reproducibility but commenting it out?

franfram:
Yeah, commenting it out sounds good!

PR #309, opened 2023-04-01
remove unnecessary cell

KhoaHoangViet:
Remove unnecessary cell (cell 30), which will do exactly what cell 29 is doing.


simonpfish:
Yup this checks out. Thank you for the contribution!

PR #310, opened 2023-04-01
add comparison benchmark for embedding models

AlexZhangji:
Massive Text Embedding Benchmark (MTEB) Leaderboard: https://huggingface.co/spaces/mteb/leaderboard


ted-at-openai (approved):
Thanks!

PR #321, opened 2023-04-07
Add getting started with AnalyticDB distributed vector database example

wangxuqi:
Hi there,

I'm excited to open this PR to add an example of support for using a fully Postgres syntax compatible database 'AnalyticDB' as a datastore. AnalyticDB is a distributed Cloud-Native vector database. It works better when data comes to large scale. 

It also support absolute out-of-box experience allow to scale into billions of data vectors processing with rich features including indexing algorithms, structured & non-structured data features, realtime update, distance metrics, scalar filtering, time travel searches etc.  It is a good choice for vector storage.


ted-at-openai:
Thanks for the contribution!

PR #322, opened 2023-04-07
Added or link.startswith("tel:") in the condition to exclude tel: links from clean links.

khapaev:
Updated the get_domain_hyperlinks function to include handling of tel: links in addition to mailto: links, to exclude them from the clean links list.


ted-at-openai:
Fantastic. Thanks for the contribution!

Issue #319, opened 2023-04-06
`wget` in GPT4_Retrieval_Augmentation.ipynb example does not download recursively


neilobremski:
The following command in `examples/vector_databases/pinecone/GPT4_Retrieval_Augmentation.ipynb` is supposed to recursively download all HTML files recursively for sample data:

```
!wget -r -A.html -P rtdocs https://langchain.readthedocs.io/en/latest/
```

However, it is only downloading the one `index.html`. I tried removing the `-A.html` and even adding `--level 10` but no luck. This happens both in the notebook and both locally when I try to run it on my machine.

ted-at-openai:
FYI @jamescalam

jamescalam:
@neilobremski @ted-at-openai I submitted a PR to fix this above. The LangChain docs have moved, so running the old command just retrieves the single `index.html` redirect page. If you update the command to:

```
!wget -r -A.html -P rtdocs https://python.langchain.com/en/latest/
```

It will work as intended!

jamescalam:
@ted-at-openai @neilobremski it should be fine to close this issue now

PR #326, opened 2023-04-10
update wget to new langchain docs url

jamescalam:
LangChain docs url has changed so the !wget in GPT4_Retrieval_Augmentation.ipynb no longer works as intended, as raised in #319 — this PR fixes that.


ted-at-openai:
Merged. Thank you!

PR #340, opened 2023-04-14
Update setup section to install openai module

roncrivera:
The openai module is not part of the setup instructions and as such the subsequent imports fail.

This PR updates the setup section to include the installation of the openai python module.


ted-at-openai:
I think it's ok for our code examples to assume openai has been installed. Hopefully folks can figure it out when they get an error. I'll close for now, but let me know if you disagree.

PR #342, opened 2023-04-14
Create SECURITY.md

iwordz:


ted-at-openai:
Seems like a good idea to add, but don't think this template is worth merging in its current state.

PR #343, opened 2023-04-14
Update README.md to remove older notebooks from recently added section

logankilpatrick:


ted-at-openai:
Whoops, just did the exact same thing in #345. Great minds think alike. :)

PR #350, opened 2023-04-15
Update the section token count from 1,000 to 1,600

eugeneyan:
This resolves the mismatch between the values in markdown cell and code cell which may confuse readers.

(BTW, between this and the downstream Q&A example, I believe this notebook is a far more valuable example.)


ted-at-openai:
Great catch - thank you for taking the time to fix it for others!

PR #372, opened 2023-04-25
Fix File Q&A build error due to removed way of assigning options to the Formidable instance

jerichosy:
Assigning a value through `form.maxFileSize` directly no longer works and will result in `Type error: Property 'maxFileSize' does not exist on type 'IncomingForm'.` during compilation when doing `npm run build`.

Instead, pass options to the function/constructor as recommended by the [Formidable docs](https://github.com/node-formidable/formidable/blob/e5e25d2fd9a018c14d7f0014f0a9b8493892b68b/README.md#formidable--incomingform).


swearer23:
so anyone to review this PR?

simonpfish (approved):

simonpfish:
Thanks for the contribution @jerichosy!

PR #378, opened 2023-04-28
Update Question_answering_using_embeddings.ipynb

acarter881:
Grammatical edit


ted-at-openai:
Will close without merging. My reasoning is that this makes the sentence structure less parallel. Would go from best/worst to best/worse.

PR #391, opened 2023-05-03
Fix Redis supported client list

joeywhelan:
Corrected the Redis client list supporting Vector Search.  @tylerhutcherson


ted-at-openai:
I'll trust this is good. Thanks!

PR #402, opened 2023-05-09
Adding SingleStoreDB as a vector database with Python notebook

arno756:
Hello,

We would like to add SingleStoreDB as an example for vector database: https://github.com/openai/openai-cookbook/tree/main/examples/vector_databases . I have included a notebook using Wikipedia dataset and working with ADA and GPT for embedding creation and getting a QA example.

Thank you,
Arnaud Comet
Director of Product


arno756:
@logankilpatrick any feedback about our proposal?

ted-at-openai:
Thanks for the contribution!

PR #408, opened 2023-05-10
Add Redis JSON w/VSS examples

joeywhelan:
These examples show how to use the Redis JSON data structure with OpenAI embeddings for VSS.


tylerhutcherson:
cc @colin-jarvis this one is good to go from our end. Thanks @joeywhelan !

examples/vector_databases/redis/redisjson/redisjson.ipynb
colin-openai, line 1: Same comment as for the other one, can you link to the relevant parts of your documentation so they can take this further if they need to?

colin-openai, line 1: Rather than verbally telling them this, why don't you print the articles plus their scores? Returning only the score doesn't prove the VSS is working. You can print the content to show we've gotten the most relevant stuff, and if you need the original doc to prove its relevant then print that too.

colin-openai, line 1: Why return only the score? I would show the content as well so they can verify it is working.

colin-openai, line 1: Same comment as the VSS example, please return the content as well in an easily readable format. If you include a simple function to unpack the payload so its easy to read into a dataframe or iterate through, then even better.

examples/vector_databases/redis/redisjson/requirments.txt
colin-openai, line 1: Requirements file has tons of dependencies, are these all needed? Try using pipreqsnb to prune to only the needed libraries.

colin-openai (changes requested):
A few changes in here, mainly around showing the content when demonstrating the similarity search. Once these are done we'll be good to merge.

examples/vector_databases/redis/redisjson/redisjson.ipynb
joeywhelan: Yes, added doc links to latest push.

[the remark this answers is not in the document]
joeywhelan: Yes, added this code to latest push.

joeywhelan: Agreed, added logic to print both.

joeywhelan: Agreed, added logic to print both.

joeywhelan: Agreed, added logic to print both in a loop.

examples/vector_databases/redis/redisjson/requirments.txt
joeywhelan: Agreed, corrected this in latest push.

PR #410, opened 2023-05-10
add simpler parallel api request script + rename robust version so th…

joe-at-openai:
Adding a simple example of how to run a parallel API request (following https://gist.github.com/neubig/80de662fb3e225c18172ec218be4917a). Also renaming the more robust parallel API request script so they will both sit side by side alphabetically.


ted-at-openai:
Won't this simple version potentially hit rate limit errors?

joe-at-openai:
Yes definitely. At best it's a quick and dirty script useful for small scale things. Once people hit rate limits they would have to switch to the robust one.

simonpfish:
Given that the purpose of the parallel script is to allow massive batch processing of requests, and not hitting API rate limits is a critical element of this, I'll be closing this PR for now. But thank you for your contribution nonetheless!

Something we could revisit is writing a notebook about parallel calls were we slowly build up to this script, and include simpler examples there.

PR #411, opened 2023-05-11
Added kusto vector database sample

anshulsharmas:
Added support for Azure Data Explorer aka Kusto to be used as a vector database for OpenAI embeddings.


ted-at-openai:
Thanks for the contribution!

Issue #418, opened 2023-05-12
[PROBLEM] Retrieval Augmentation for GPT-4 using Pinecone has a undefined variable 'data'


jimmiemunyi:
**Identify the file to be fixed**
examples/vector_databases/pinecone/GPT4_Retrieval_Augmentation.ipynb

**Describe the problem**
From the 8th code cell onwards, there is a reference to a variable called `data` which was not defined earlier,
hence the notebook cannot be run from there. It looks like the cell was deleted before committing the file.

**Describe a solution**
Create a cell before that defines how the `data` variable should be defined.

**Screenshots**
![image](https://github.com/openai/openai-cookbook/assets/38419676/5f0b896c-4607-4297-b122-919f1eb532b2)

ted-at-openai:
FYI @jamescalam

ted-at-openai:
Fix merged: https://github.com/openai/openai-cookbook/pull/424

Will close issue, assuming it's resolved.

PR #424, opened 2023-05-15
added data creation loop

jamescalam:
Fix to issue #418 — adds a missing data formatting loop that creates `data`


ted-at-openai:
Thanks - merging now.

PR #430, opened 2023-05-18
Update transformers.py

Optimized-Coder:
Add explanative docstring to handle_file_string function and move docstrings to below function definition as per pep 257 specification


ted-at-openai:
Thanks! Looks like a no-op in terms of functionality, so will merge without personally testing. Let me know if mistaken.

PR #433, opened 2023-05-19
small change to move every api call inside while loop

ancri:
Currently there's an `openai.Embedding.create` call outside the `while` loop, and then the call is replicated inside of the exception-handling loop. This seems superfluous, so simplifying it slightly.


ted-at-openai:
@jamescalam, look good to you?

ted-at-openai:
Looking more closely, this doesn't look right. It deletes the while loop. Now the done variable is never used and some embeddings may never be calculated. Let me know if I've misunderstood. Otherwise, I'll close.

ancri:
Thanks, you're right, apologies. Opened another PR where this is fixed.

PR #437, opened 2023-05-19
updates unit test writing example to use gpt-3.5-turbo

ted-at-openai:


knas12000 (commented):
I'm learning, please have patience.

PR #446, opened 2023-05-23
Removing reference to removed azure fine tuning notebook.

glecaros:


ted-at-openai (approved):
Thanks for the fast follow!

PR #447, opened 2023-05-23
Fix mutable default argument in APIRequest class

isabelaconstantin:
This change prevents `result` to be shared among all instances of `APIRequest` (see this common [gotcha](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments))
Otherwise, with the current initialisation, the `result` instance variable will gradually accumulate all errors encountered so far in the previous API Requests.


ted-at-openai (approved):

isabelaconstantin:
Thanks!

PR #448, opened 2023-05-23
Add ability to store row-level metadata

domenicomanna:
This metadata field allows non API parameters to be stored which could be useful for when the results need to be associated with an external source. For example, if the user has several rows of text data within a database, and they want to calculate embeddings for each row, they can now use the metadata field to save the row id so that when they parse the results back, they know which embedding corresponds to which database row.


ted-at-openai:
Good idea, thanks for the contribution. I considered something like this before, but omitted it since the `input` field itself is going to be the most reliable key. Still, I could see `metadata` helping if the metadata is more hashable or if there are duplicate inputs that you want to more easily match up.

PR #452, opened 2023-05-24
fix label 'contradiction' dropped bug

dhrim:
in function process_input_data() label "contradiction" was missed when filtering "entailment" or "contradiction".
So all data was only of label "entailment"

I just added "contradition".


ted-at-openai:
I don't think this is a bug - the point of the section is to show how you can start with only positives and then generate synthetic negatives.

dhrim:
got it

PR #453, opened 2023-05-24
Update web-qa.py

jpa-rocha:
previously it was:
row['n_tokens'] + 4 
Error output was:
raise KeyError(key) from err
KeyError: 'n_tokens'


simonpfish:
Can't reproduce. I think `n_tokens` gets computed here: https://github.com/openai/openai-cookbook/blob/791e96ba06429582447df6c25e4872ec124f78d3/apps/web-crawl-q-and-a/web-qa.py#L211

PR #467, opened 2023-05-28
Enhancements and Refactoring of Python Code Extraction Methods

eli64s:
#### Description
This pull request introduces enhancements and refactoring to the [Code_search.ipynb](https://github.com/openai/openai-cookbook/blob/main/examples/Code_search.ipynb) notebook, which contains methods for extracting Python functions and generating text embeddings. The proposed modifications not only make the code more efficient and user-friendly, but also ease the process of future maintenance. 

#### Included Updates

1. **Normalization of File Paths:** In the current implementation, file path normalization is done using `string.replace()`. This approach produces inconsistent results and may potentially replace substrings not part of the root directory. To address this issue, I've updated the code to utilize the `relative_to()` method from the `pathlib.Path` module. This method accurately computes relative paths, considering the file structure, and ensures correct results in scenarios when the base directory name appears multiple times in the path. Please refer to the example below for a better understanding of the differences between the two approaches.

    ```python
    import pandas as pd
    from pathlib import Path
    
    data = {'file_path': [
        'repo/main/src/file1_copy/other/repo/main/src/file1',
        'repo/main/src/file1_copy/file1',
    ]}
    df = pd.DataFrame(data)
    
    # Approach 1: Path.relative_to()
    root_dir = Path('repo/main/src')
    df['Path.relative_to()'] = df['file_path'].map(lambda x: Path(x).relative_to(root_dir))
    
    # Approach 2: string.replace()
    root_dir = 'repo/main/src'
    df['str.replace()'] = df['file_path'].apply(lambda x: x.replace(root_dir, ''))
    ```
    
    |   | file_path                                          | Path.relative_to()               | string.replace()            |
    |--:|:---------------------------------------------------|:-------------------------------------|:-------------------------|
    | 0 | repo/main/src/file1_copy/other/repo/main/src/file1 | file1_copy/other/repo/main/src/file1 |  **`/file1_copy/other//file1`** |
    | 1 | repo/main/src/file1_copy/file1                     | file1_copy/file1                     | /file1_copy/file1        |

2. **Capture `async def`:** Updated the file searching logic to extract both regular `def` and asynchronous `async def` methods. This modification enables a more robust extraction of Python functions and ensures that async methods are not excluded from the analysis. 

3. **Refactor `get_functions`:** Refactored function to handle files using a context manager to improve safety and resiliency.

4. **Refactor `get_until_no_space`:** Update logic to prevent potential index out of range errors.

5. **Improve Directory Search:** Update code to use *pathlib.Path.glob()* to search for files vs. the original *os.walk()* and *glob()* methods. The *os.walk()* method traverses the directory tree recursively, generating a tuple for each directory it encounters. The *pathlib.Path.glob()* method performs the file search directly, without generating intermediate results. This can lead to improved performance, as the search is more efficient and consumes less memory.

6. **Implement `extract_functions_from_repo`:** Encompasses the logic of code file function extraction and printing.

These changes collectively enhance the functionality and maintainability of the script, providing better support for future development and analysis tasks involving the [openai-cookbook](https://github.com/openai/openai-cookbook) repository.

Best,
Eli


ted-at-openai:
I will try to review this week. Thanks for the detailed and high-quality contribution!

examples/Code_search.ipynb
```diff
@@ -243,121 +148,27 @@
   },
```
ted-at-openai, line 148: Now that filepaths are filepaths instead of strings, I believe `r[1].filepath` on line 141 needs to be wrapped in `str()` so it can print.

Otherwise, everything ran great for me. Thanks!

ted-at-openai (changes requested):

examples/Code_search.ipynb
eli64s: Updated in new commit, thanks.

ted-at-openai (approved):
Apologies for taking so long. Approving and merging. Thank you!

ted-at-openai:
By the way, really appreciate you taking the time to describe and document your improvements. Always love to see it. :)

PR #472, opened 2023-06-02
add arthur shield to external libraries section

ted-at-openai:


apwenchel:
Thanks @ted-at-openai !

PR #481, opened 2023-06-08
Add example notebook for financial document analysis with LlamaIndex

Disiok:
# Summary
* Add example notebook for financial document analysis with LlamaIndex
* Add 2 10-K documents (PDFs) into `data/examples/data/10k`
> This may have undesired effect in making repo bigger and thus `git clone` slower. Let me know if this is a big concern. Alternatives could be to use `git-lfs` to store this, or just point to the same files hosted at llama-index GitHub repo.


examples/financial_document_analysis_with_llamaindex.ipynb
```diff
@@ -0,0 +1,554 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d1062e24-6177-459a-94e0-8e77b3e0859b",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "!pip install llama-index"
```
colin-openai, line 92: Can you add pypdf and Langchain here - I got an error on the SimpleDirectoryReader because of no pypdf

examples/financial_document_analysis_with_llamaindex.ipynb
```diff
@@ -0,0 +1,554 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c743f504-f28c-4802-89b6-ad152b74b0eb",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "Before we start, we can configure the LLM provider and model that will power our RAG system.  \n",
+    "Here, we pick *text-davinci-003* from OpenAI, allow unlimited output tokens.  "
```
colin-openai, line 129: Is there a particular reason we use the Completions model rather than the Chat models? Given our new ones are Chat-based, it'd be great to refactor this to use them if possible.

If you don't currently support this we can push as-is and make a subsequent notebook.

examples/financial_document_analysis_with_llamaindex.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fd122d0d-2da6-4f46-aa2a-8a0049ad8694",
+   "metadata": {},
+   "source": [
+    "Now, we can build an (in-memory) `VectorStoreIndex` over the documents that we've loaded.  \n",
+    "\n",
+    "> Note: this operation might take a while to run, since it calls OpenAI API for computing vector embedding over document chunks."
```
colin-openai, line 230: Does it do this in parallel or in sequence - if in sequence, can it done in parallel to reduce execution time? Also, what happens if some of the embeddings fail?

colin-openai (changes requested):
Generally looks great, but would be better using ChatCompletions. If possible to refactor, that would be great before merging.

I'd also like to move this into a sub-directory, just agreeing naming convention with the team before we do.

examples/financial_document_analysis_with_llamaindex.ipynb
Disiok: sure, we can swap to chat models. Right now we use langchain LLM abstractions so it's a one-line change.

examples/financial_document_analysis_with_llamaindex.ipynb
Disiok: good catch. will do!

examples/financial_document_analysis_with_llamaindex.ipynb
Disiok: It calls the endpoint in batches. And I think if some embeddings fail, the whole operate would fail. Definitely should add some better error catching here.

examples/financial_document_analysis_with_llamaindex.ipynb
Disiok: Hey @colin-openai , just tried to update the notebook to using the latest `gpt-3.5-turbo-0613` model, but was getting some qualitatively worse answers (model refusing to give an answer, saying "Based on the provided context information, it is not explicitly mentioned...."). 

I think we might need to do some prompt tuning to update for the new model. Maybe we can just land this as is, and then update this after we do some more experiments?

examples/financial_document_analysis_with_llamaindex.ipynb
Disiok: Added `pypdf`.

`langchain` is a dependency of `llama-index`, so I think it's unnecessary here.

colin-openai (approved):
Happy with changes, merging

PR #483, opened 2023-06-08
Initial commit of reranking with cross-encoder cookbook

colin-openai:


juntang-zhuang:
Hi, it looks good to me in general. One corner case I'm thinking is that the model might output the next most probable token not in ['yes','no'] (I expect this to happen at an extremely low frequency). In this case, we might need to take out the logits for ['yes','no'] then perform a softmax as post-processing.

ted-at-openai:
> Hi, it looks good to me in general. One corner case I'm thinking is that the model might output the next most probable token not in ['yes','no'] (I expect this to happen at an extremely low frequency). In this case, we might need to take out the logits for ['yes','no'] then perform a softmax as post-processing.

Two options here, since you cannot grab the full distribution:
- Use logit_bias to massively upweight the ' yes' and ' no' tokens (better if you need both tokens)
- Use `echo=True` to always get the logprob for ` yes` (better if you just need ` yes`)

colin-openai:
@ted-at-openai @juntang-zhuang updated with logit bias for Yes and No, seems to work well. Let me know if there is any other feedback, otherwise ready to merge.

juntang-zhuang (approved):
LGTM, thanks for updating!

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7f30b8b2",
+   "metadata": {},
+   "source": [
+    "# Reranking search results with cross-encoders\n",
```
ted-at-openai, line 8: small suggestion: could name it Search reranking with cross-encoders, just so it shows up next to the other search examples alphabetically

(fine either way)

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+       "([5297], [2949])"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "tokens = ['Yes','No']\n",
```
ted-at-openai, line 194: include leading spaces in the tokens so they match the few-shot prompt (e.g. ` Yes` instead of `Yes`)

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d5727678",
+   "metadata": {},
+   "source": [
+    "## Cross-encoder\n",
+    "\n",
+    "We'll create a cross-encoder using the ```Completions``` endpoint - the key factors to consider here are:\n",
+    "- Make your examples domain-specific - the strength of cross-encoders comes when you fine-tune them to your domain.\n",
```
ted-at-openai, line 166: Suggestion: replace 'fine-tune' with 'tailor' or 'customize' or 'adapt' or something, so that people don't think this requires model fine-tuning

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+    "- There is a trade-off between how many potential examples to re-rank vs. processing speed. Consider batching and parallel processing cross-encoder requests to aid user experience.\n",
```
ted-at-openai, line 167: 'aid user experience' sounds abstract. Maybe: 'process them faster'?

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "tokens = ['Yes','No']\n",
+    "tokenizer = tiktoken.get_encoding(\"gpt2\")\n",
```
ted-at-openai, line 195: could use `encoding_for_model` to show people there's a way to get encodings without knowing that `gpt2` corresponds to `text-davinci-003`

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+    "\n",
+    "We'll create a cross-encoder using the ```Completions``` endpoint - the key factors to consider here are:\n",
+    "- Make your examples domain-specific - the strength of cross-encoders comes when you fine-tune them to your domain.\n",
+    "- There is a trade-off between how many potential examples to re-rank vs. processing speed. Consider batching and parallel processing cross-encoder requests to aid user experience.\n",
+    "\n",
+    "The steps here are:\n",
+    "- Build a prompt to assess relevance and provide few-shot examples to tune it to your domain\n",
+    "- Return the classification of yes/no as well as the ```logprobs```\n",
+    "- Rerank the results by the ```logprobs``` keyed on Yes\n",
+    "- Add a ```logit bias``` for the tokens for Yes and No to decrease the likelihood of any other tokens occurring"
```
ted-at-openai, line 173: Should this go above the prior two bullets? Feels out of order, temporally

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+    "content = result_list[0][\"title\"] + \": \" + result_list[0][\"summary\"]\n",
+    "\n",
+    "# Set logprobs to 1 so we return the most likely token with the likelihood\n",
+    "response = openai.Completion.create(\n",
+    "    model=\"text-davinci-003\",\n",
+    "    prompt=prompt.format(query=query, document=content),\n",
+    "    temperature=0,\n",
+    "    logprobs=1,\n",
+    "    logit_bias={\n",
+    "            5297: 1,\n",
```
ted-at-openai, line 287: can go higher than 1 for extra reliability (e.g., 100)

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "prompt = '''\n",
+    "You are an Assistant responsible for helping detect whether the retrieved document is relevant to the query. For a given input, you need to output a single token: \"Yes\" or \"No\" indicating the retrieved document is relevant to the query.\n",
+    "\n",
+    "Examples:\n",
+    "\"\"\"\n",
+    "Query: How to plant a tree?\n",
+    "Retrieved document: Cars were invented in 1886, when German inventor Carl Benz patented his Benz Patent-Motorwagen.[3][4][5] Cars became widely available during the 20th century. One of the first cars affordable by the masses was the 1908 Model T, an American car manufactured by the Ford Motor Company. Cars were rapidly adopted in the US, where they replaced horse-drawn carriages.[6] In Europe and other parts of the world, demand for automobiles did not increase until after World War II.[7] The car is considered an essential part of the developed economy.\n",
```
ted-at-openai, line 213: optional: you can potentially wrap these strings in delimiters (e.g., `"""`). Otherwise it might get less reliable on documents or queries that have newlines and colons within them.

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+   "cell_type": "code",
+   "execution_count": 7,
+   "id": "4fdf8c11",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "prompt = '''\n",
+    "You are an Assistant responsible for helping detect whether the retrieved document is relevant to the query. For a given input, you need to output a single token: \"Yes\" or \"No\" indicating the retrieved document is relevant to the query.\n",
+    "\n",
+    "Examples:\n",
```
ted-at-openai, line 210: I highly, highly recommend not telling the model that these are examples and that the final query is different. Best practice is to format the final query exactly as the examples to minimize the odds of treating it differently.

So concretely, I would:
- Remove the Examples label
- Remove the End of Examples label
- Remove the Begin label
- And just make the final query look like a half-written example

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "id": "4fdf8c11",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "prompt = '''\n",
+    "You are an Assistant responsible for helping detect whether the retrieved document is relevant to the query. For a given input, you need to output a single token: \"Yes\" or \"No\" indicating the retrieved document is relevant to the query.\n",
```
ted-at-openai, line 208: Not sure if telling the model it is an assistant will help. Could plausibly hurt performance a little, if it thinks assistants don't operate at peak performance. Not a big deal either way.

ted-at-openai: An alternative could just be triplets of:
- Query:
- Document:
- Relevant:[ Yes/ No]

ted-at-openai: One advantages of this^ alternative is that `Relevant to the query` won't get mixed in with the retrieved document text. Right now it's just separated with a newline, which isn't a lot if the document also has newlines.

examples/Reranking_search_results_with_cross-encoders.ipynb
ted-at-openai: and if you want 100% reliability, you can literally append ` Yes` to the end and then just do echo=True and take the last logprob. Guarantees you always get a logprob for ` Yes`. No need to make the change unless you want to.

ted-at-openai: One problem with appending is that you'll get the raw logprob for ` Yes`, which won't just be the ratio of ` Yes` to ` No` but ` Yes` to every other possible token. One nice benefit of the ` Yes` & ` No` upweighting is that you effectively end up with the ratio between them, so you aren't hurt by cases where the model wants to respond with a third thing.

Bottom line: I recommend keeping as is. (Plus, all of these considerations are minor for a task like this which ought to be quite reliable.)

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "id": "753cd363",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "content = result_list[0][\"title\"] + \": \" + result_list[0][\"summary\"]\n",
+    "\n",
+    "# Set logprobs to 1 so we return the most likely token with the likelihood\n",
```
ted-at-openai, line 280: nit: maybe rephrase this comment? wasn't sure what it was saying when I first read it

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+    "\n",
+    "Query: {query}\n",
+    "Retrieved document: {document}\n",
+    "Relevant to the query?\n",
+    "Assistant:\n",
+    "'''\n",
+    "\n",
+    "\n",
+    "@retry(wait=wait_random_exponential(min=1, max=40), stop=stop_after_attempt(3))\n",
+    "def rerank_document(query, document, output_list):\n",
```
ted-at-openai, line 249: not a big deal, but one way to do this function would be to have it return the output list and then do the appending outside of the function. I tend to prefer functions that return a value rather than mutate the state of an input (feels easier for readers to follow), but either way works.

ted-at-openai, line 249: Would it make more sense to call this function `document_relevance` or something like that? It's not actually doing any reranking, right? It's just returning logprobs of the relevance tokens?

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+    }
+   ],
+   "source": [
+    "output_df = pd.DataFrame(\n",
+    "    output_list, columns=[\"query\", \"document\", \"prediction\", \"logprobs\"]\n",
+    ").reset_index()\n",
+    "# Use exp() to convert logprobs into probability\n",
+    "output_df[\"probability\"] = output_df[\"logprobs\"].apply(exp)\n",
+    "# Reorder based on likelihood of being Yes\n",
+    "output_df[\"cleaned_probability\"] = output_df.apply(\n",
```
ted-at-openai, line 540: `cleaned_probability` confused me at first - maybe call it `yes_probability` instead? I think that would be clearer to someone scanning the table and wondering why `probability` is dirty.

Or, an alternative way of doing this could be to:
- Return the top 2 logprobs instead of top 1
- Filter to yes rows

Then you don't need the complexity of a lambda function and two differently named probability columns

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+   "source": [
+    "reranked_df['document'][0]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f372d311",
+   "metadata": {},
+   "source": [
+    "## Consolidation\n",
```
ted-at-openai, line 776: What does Consolidation mean?

Was expecting this to say Conclusion :)

ted-at-openai: In any case, there might be some room to add explanation of how this compares and why it might be beneficial to use. Right now it just kind of abruptly ends without any discussion of how much better these results are than an alternative.

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "7f30b8b2",
+   "metadata": {},
+   "source": [
+    "# Reranking search results with cross-encoders\n",
+    "\n",
+    "This notebook takes you through examples of using a cross-encoder to re-rank search results.\n",
+    "\n",
+    "This is a common use case with our customers, where you've implemented semantic search using embeddings (produced using a [Bi-encoder](https://www.sbert.net/examples/applications/retrieve_rerank/README.html#retrieval-bi-encoder)) but the results are not exactly what you need. A possible cause is that there is some business rule you can use to rerank the documents such as how recent or how popular a document is. \n",
```
ted-at-openai, line 12: nit: capitalization is inconsistent - Bi-encoder is capitalized once, but then cross-encoder is lowercase.

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+    "\n",
+    "However, often there are subtle domain-specific rules that help determine relevancy, and this is where a cross-encoder can be useful. Cross-encoders are more accurate than bi-encoders but they don't scale well, so using them to re-order a shortened list returned by semantic search is the ideal use case.\n",
+    "\n",
+    "We'll use text-davinci-003 with logprobs to build a cross-encoder here.\n",
```
ted-at-openai, line 16: optional: could be worth emphasizing that GPT's advantage of being a cross-encoder is its language understanding

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+   "cell_type": "markdown",
+   "id": "7f30b8b2",
+   "metadata": {},
+   "source": [
+    "# Reranking search results with cross-encoders\n",
+    "\n",
+    "This notebook takes you through examples of using a cross-encoder to re-rank search results.\n",
+    "\n",
+    "This is a common use case with our customers, where you've implemented semantic search using embeddings (produced using a [Bi-encoder](https://www.sbert.net/examples/applications/retrieve_rerank/README.html#retrieval-bi-encoder)) but the results are not exactly what you need. A possible cause is that there is some business rule you can use to rerank the documents such as how recent or how popular a document is. \n",
+    "\n",
```
ted-at-openai, line 13: I think it's worth adding a paragraph to introduce what a cross-encoder is.

Here is a suggestion:

"
Consider a search task with D documents and Q queries.

The brute force approach of computing every pairwise relevance is expensive; its cost scales as `D * Q`. This is known as cross-encoding.

A faster approach is embeddings-based search, in which an embedding is computed once for each document and query; its cost scales as `D + Q`. The embeddings are then re-used multiple times to cheaply compute cosine similarity pairs. This is known as bi-encoding.

Although embeddings-based search is faster, the quality can be worse. To get the best of both, one common approach is to use embeddings (bi-encoder) to cheaply identify top candidates, and then use GPT (cross-encoder) to expensively re-rank those top candidates. The cost of this hybrid approach scales as `(D + Q) * cost of embedding + (N * Q) * cost of re-ranking`, where N is the number of candidates re-ranked.

examples/Reranking_search_results_with_cross-encoders.ipynb
```diff
@@ -0,0 +1,805 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "fdada886",
+   "metadata": {},
+   "source": [
+    "## Search\n",
+    "\n",
+    "We'll use the arXiv search service for this example, but this step could be performed by any search service you have. The key item to consider is over-fetching slightly to capture all the potentially relevant documents, before re-sorting them.\n",
+    "\n",
+    "To illustrate our point we'll sort by \"SubmittedDate\" rather than \"Relevance\" so we get some irrelevant results in the top 5."
```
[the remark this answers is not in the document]
colin-openai: Removing this text as using SubmittedDate resulted in no remotely relevant docs most of the time

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Got it, I'll keep as-is for now, but will ask you about this implementation to understand it better.

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Took this approach and reworded

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Refactored this one to return a tuple which we can append outside.

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: I also tested this implementation and still got a fair number of No's even with a big logit_bias on " Yes" and none on No. Might be worth a discussion.

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Updated wording here, it was confusing

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Updated to be `document_relevance`

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Changed to `yes_probability` - tried top 2 logprobs but often got confounding ones even with `logit_bias` turned up for Yes and No i.e. I would get `No` and ` No` and the top two values.

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: I've added this example, though there is some overlap with the second and third paragraphs. Open to your thoughts, we can replace with the above but I want to retain some of the simplicity of the first couple of paragraphs.

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Good point, have added a sentence after the mention of `text-davinci-003`

examples/Reranking_search_results_with_cross-encoders.ipynb
colin-openai: Right, I sometimes say Consolidation but it is harder to understand, updated to Conclusion and gave a few suggestions of use cases and next steps.

colin-openai:
@ted-at-openai thanks for reviewing, I've left open the ones that need an extra check on how I've responded - once you're happy I'll merge.

examples/Reranking_search_results_with_cross-encoders.ipynb
ted-at-openai: Oh, interesting. Maybe the boost wasn't high enough for ` Yes` to outweigh `No`?

ted-at-openai:
Looks great. Only two final recommendations:
- Use a big logit_bias (e.g., 100 instead of 1)
- Rerun all the cells at the end so they're in numerical order (without any distracting jumps from 5 to 30 where someone wonders what the author was doing for those 25 steps in between :) )

PR #497, opened 2023-06-13
Joe/generate function arguments

joe-at-openai:


ted-at-openai:
Should I close this PR? (Assuming it was merged in another a couple weeks ago)

PR #498, opened 2023-06-13
Fix ask_database in call functions with chat model example

pycui:
Fix function description, hopefully make it more clear for readers and model.


colin-openai:
Looks good and makes sense. Merging.

PR #519, opened 2023-06-15
Add search augmentation notebook

simonpfish:
Notebook intro: 
> Searching for relevant information can sometimes be like looking for a needle in a haystack, but don’t despair, GPTs can actually do a lot of this work for us. In this guide we explore a way to augment existing search systems with various AI techniques, helping us sift through the noise.


examples/search_augmentation.ipynb
```diff
@@ -0,0 +1,472 @@
...
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Number of articles: 185\n",
+      "Top 5 articles: \n",
```
ted-at-openai, line 199: I got a little confused here: are these the top 5 of query #1? did the prior step just concatenate our relevancy-ranked lists together?

examples/search_augmentation.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "    text = completion.choices[0].message.content\n",
+    "    parsed = json.loads(text)\n",
+    "\n",
+    "    return parsed\n",
+    "\n",
+    "\n",
+    "def embedding(input):\n",
+    "    response = openai.Embedding.create(\n",
+    "        model=\"text-embedding-ada-002\", input=input)\n",
+    "    return [data.embedding for data in response.data]\n"
```
ted-at-openai, line 87: could optionally name the function embeddings rather than embedding, given that it always returns a list (makes me think plural rather than singular)

examples/search_augmentation.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "import dotenv\n",
+    "import requests\n",
+    "import json\n",
+    "from datetime import date, timedelta\n",
+    "from numpy import dot\n",
+    "from IPython import display\n",
+    "\n",
+    "\n",
+    "# Load environment variables\n",
+    "dotenv.load_dotenv()\n",
```
ted-at-openai, line 62: is this necessary? I can read my openai key with just os.getenv()

examples/search_augmentation.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "        f\"{article['title']} {article['description']} {article['content'][0:100]}\"\n",
+    "        for article in articles\n",
+    "    ]\n",
+    ")\n",
+    "\n",
+    "# Calculate cosine similarity\n",
+    "cosine_similarities = []\n",
+    "for article_embedding in article_embeddings:\n",
+    "    cosine_similarities.append(\n",
+    "        dot(hypothetical_answer_embedding, article_embedding)[0])\n",
```
ted-at-openai, line 323: this line looks strange with a closing paren but no opening paren; when I format with black, it puts this all on one line

examples/search_augmentation.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    }
+   ],
+   "source": [
+    "formatted_top_results = [\n",
+    "    {\n",
+    "        \"title\": article[\"title\"],\n",
+    "        \"description\": article[\"description\"],\n",
+    "        \"url\": article[\"url\"],\n",
+    "    }\n",
+    "    for article, _score in sorted_articles[0:50]\n",
```
ted-at-openai, line 423: could be interesting to do the top 5 instead of top 50 so that the user can personally judge the model (otherwise they're just taking this answer on faith). could also put the raw and re-ranked answers side-by-side to see the improvement caused by re-ranking.

ted-at-openai (changes requested):

examples/search_augmentation.ipynb
ted-at-openai, line 323: also worth commenting why dot product works as cosine similarity, since that's not generally true

examples/search_augmentation.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "\n",
+    "Let's dive into it! We will use Twitter as an example domain to search over.\n",
+    "\n",
+    "## Setup\n",
+    "\n",
+    "Once you have your keys, you can set them as environment variables in your`.env` file in the same directory as this notebook. The `.env` file should look like this:\n",
+    "\n",
+    "```\n",
+    "NEWS_API_KEY=your_api_key\n",
+    "OPENAI_API_KEY=your_openai_api_key\n",
```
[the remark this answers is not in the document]
simonpfish: I rewrote this section making a bit clearer on how to handle and configure the keys

ted-at-openai (approved):
One more possible small tweak - could rename the image file

PR #535, opened 2023-06-22
User_and_product_embeddings.ipynb Dataset update

brunobelloni:
dataset is slightly smaller compared to the original example, but it works for educational purposes


examples/User_and_product_embeddings.ipynb
```diff
@@ -1,13 +1,15 @@
 {
  "cells": [
   {
-   "cell_type": "markdown",
-   "metadata": {},
+   "cell_type": "raw",
```
ted-at-openai, line 4: this should stay as markdown

ted-at-openai:
Thanks for the contribution.

Code seems to run, but text is now out of date.

Also, it looks like the conclusion (trend in scores) doesn't appear to hold in the smaller sample.

So I think I'll close this and we'll work on providing better, cleaned up instructions for the full dataset. 

Thank you nonetheless.

PR #538, opened 2023-06-22
Fix wget download updated parameters

michaelskyuan:
Execution of getting-started-with-redis-and-openai.ipynb was failing due to broken parameters in wget call in nbutils.py
@Spartee @colin-openai


colin-openai (approved):
Looks good, happy to merge

PR #543, opened 2023-06-24
consolidate Embedding.create calls into one

ancri:
@ted-at-openai apologies for removing the while loop in the previous PR. here is the version that should maintain the exact same functionality as the original, just with a bit more economic / clear implementation and no repetition of the Embedding.create call


ted-at-openai:
Thanks! Merging.

PR #545, opened 2023-06-24
Update How_to_call_functions_with_chat_models.ipynb

hack-r:
The original had "min" and "max" parameters. "max" is a real parameter - the max time, i.e. timeout. There's no "min" parameter. It should be "multiplier".

https://tenacity.readthedocs.io/en/latest/api.html

### DIFF
--- a/examples/How_to_call_functions_with_chat_models.ipynb
+++ b/examples/How_to_call_functions_with_chat_models.ipynb
@@ -81,7 +81,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "@retry(wait=wait_random_exponential(min=1, max=40), stop=stop_after_attempt(3))\n",
+    "@retry(wait=wait_random_exponential(multiplier=1, max=40), stop=stop_after_attempt(3))\n",
     "def chat_completion_request(messages, functions=None, function_call=None, model=GPT_MODEL):\n",
     "    headers = {\n",
     "        \"Content-Type\": \"application/json\",\n",


ted-at-openai:
Great catch - thank you!

ted-at-openai:
Wait, I do see a `min` parameter, both in the `tenacity` library and also the docs. It's the final argument for `classtenacity.wait.wait_random_exponential` here: https://tenacity.readthedocs.io/en/latest/api.html#tenacity.wait.wait_random_exponential

ted-at-openai:
Looks like `multiplier` is the delay on the first retry (defaulting to 1 second) and `min` is the minimum wait time, which overrides the `multiplier` and any random jitter. I assume the writer of the code meant to use `multiplier` but appears to have a similar effect.

ted-at-openai (approved):

hack-r:
OK, sorry, maybe I need to have my eyes checked. 

It gave me some error in the IDE then when I checked the docs I guess I saw `multiplier: Union[int, float] = 1, max: Union[int, float, timedelta]` with no `min` right before or after the `max` argument (but yes, I do see it at the end of the argument list now). To help fool me it seems their example usage also uses just `multiplier` and `max`.

I will have to go back and check what the original error was that I encountered, but yes, good catch.

hack-r:
Ah ,yes, your latter comment probably explains the original error I was encountering. I'll check first chance I get.

ted-at-openai:
Yep, I missed it too on first glance. I still merged the changes because I suspect `multiplier` was the intention of the author.

PR #548, opened 2023-06-26
Added num_tokens_from_messages_and_functions method

altryne:
This method is my first attempt at counting the number of tokens, including the function definitions and provided schema.

Basically copying the previous method and running it over functions definitions. 

Open question for OAI team:
- Functions names may have different token length than just messages? Should `tokens_per_name` change for functions?
- Given that `parameters` is an object, I casted it to string with json.dumps, not sure that this was correct


ted-at-openai:
Thanks for the contribution!

I will take a look this week, and check our internal codebase / fellow engineers to see how we'd want this function to count in a way that's accurate, reliable, and future-proof.

altryne:
Thanks Ted!

I'm completely fine if this whole function is discarded and rewritten, just
wanted to give some folks at least a starting point!

Thanks for taking a look, lmk if I can help test 🫡

ahmed-mohamed-sn:
@ted-at-openai I was wondering if this functionality will be added.

PR #552, opened 2023-06-27
update to token for zilliz

filip-halt:
Updating the zilliz notebooks to use a token instead of username and password. Required for it to work with current cloud.


ted-at-openai (approved):

ted-at-openai:
Merged - thanks!

PR #559, opened 2023-07-01
fix bug in function group_chunks which not reset cur_tokens

dhrim:
Reset code was missing during iteration.

I just inserted reset code in iteration.


ted-at-openai:
Thanks! Fixed in another PR.

PR #561, opened 2023-07-03
replace eval with safer literal_eval

TDulka:
When looking through the examples I noticed quite a few are using `eval`, this might be potentially unsafe if someone runs it let's say on an embeddings file from someone else, could execute arbitrary python code hidden there. I am suggesting replacement with `literal_eval`, which is also used in other examples in the repo.

Also VSCode seems to insist on making some formatting changes that I haven't figured out how to stop. If this is an issue please feel free to reject.

### DIFF
--- a/apps/web-crawl-q-and-a/web-qa.ipynb
+++ b/apps/web-crawl-q-and-a/web-qa.ipynb
@@ -1114,10 +1114,11 @@
    "source": [
     "import pandas as pd\n",
     "import numpy as np\n",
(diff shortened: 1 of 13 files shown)


ted-at-openai:
Thanks. This looks safer. Have you checked these examples? Otherwise I'll find time to check this week before merging.

ted-at-openai (approved):
Checked most files. Looks good. One big change to Atlas example that I didn't follow.

PR #575, opened 2023-07-10
simplify pretty print

adhillo:
- drop the need for the "formatted_messages" setup
-  bringing more in line with the simpler "Conversation" class used here ([link](https://github.com/openai/openai-cookbook/blob/950246dd0810470291aa9728c404a01aeab5a1e9/examples/How_to_call_functions_for_knowledge_retrieval.ipynb))


joe-at-openai (approved):
Nice improvement. Thanks!

PR #587, opened 2023-07-13
Question and Answers notebook: update the actual number of results used in generating the final answer

yoavf:
The comment mentions 20, but it actually uses 5


ted-at-openai (approved):

ted-at-openai:
Thanks!

PR #595, opened 2023-07-19
[azure] add functions notebook sample

kristapratico:


examples/azure/functions.ipynb
```diff
@@ -0,0 +1,431 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "if not use_azure_active_directory:\n",
+    "    openai.api_type = \"azure\"\n",
+    "    openai.api_key = \"\"  # Add your api key here"
```
ted-at-openai, line 109: Happy to merge.

One thing I noticed here is that we're encouraging developers to put their API keys in code. In general, I really want to discourage this, as the leak of an API key can be quite costly. Do you mind removing this and using environment variables instead?

ted-at-openai (approved):

examples/azure/functions.ipynb
kristapratico: Thanks for reviewing, that's good feedback. I updated the code to use an environment variable for API key across all the azure examples.

ted-at-openai: Awesome. Thank you!

ted-at-openai:
Merged!

PR #596, opened 2023-07-20
feat: add supabase pgvector cookbooks.

thorwebdev:
Add useful pgvector cookbooks that work with Supabase.


examples/vector_databases/supabase/llamaindex/llamaindex.ipynb
```diff
@@ -0,0 +1,189 @@
...
+    },
+    {
+      "attachments": {},
+      "cell_type": "markdown",
+      "id": "7a95a440",
+      "metadata": {},
+      "source": [
+        "# Set up OpenAI\n",
+        "\n",
+        "LlamaIndex uses OpenAI GPT-3 `text-davinci-003` model. Let's store that in an environment variable:"
```
ted-at-openai, line 71: This model is slated for deprecation (shutdown in Jan), so might not be a good one to use for a long-term example.

thorwebdev: Thanks for the review @ted-at-openai, I've went ahead and removed the llamaindex example. PTAL.

simonpfish:
Hi @thorwebdev! First of all, thank you for your contribution. Unfortunately, these notebooks fall outside of the scope of the OpenAI Cookbook, since they don't use any of our API endpoints. 

If you add OpenAI functionality, we'd be happy to take a look.

PR #612, opened 2023-07-27
Add llm-client to README.md

uripeled2:


simonpfish:
Thank you for your contribution. As llm-client doesn't have much traction currently, we believe it's not the right time to include it in our resources. However, please feel free to resubmit it at a later time.

PR #629, opened 2023-08-11
adds one function calling notebook

prestontuggle:


ted-at-openai:
Want to run this notebook from top to bottom so people can see what example outputs look like without having to get a Google place API key first?

ted-at-openai (approved):

PR #636, opened 2023-08-15
Update web-qa.py to better handle file names for domains other than openai.com

Cadence-GitHub:
It makes it easier to use this code with domains other than openai.com


simonpfish (commented):
👍 

thanks, i like making this more general. 

how did you test that this works correctly?

simonpfish:
Closing due to inactivity. @logankilpatrick might be worth taking a look at integrating this in the future.

PR #648, opened 2023-08-22
Fix UTF-8 encoding in Chat_finetuning_data_prep.ipynb

recordcrash:
The following fixes checking .jsonl finetuning files that contain UTF-8 characters, specifically the error "UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 737: character maps to <undefined".

I also removed an unused "os" import.


simonpfish (approved):
thank you for the fix :)

PR #650, opened 2023-08-22
Update README.md to add fting guide

logankilpatrick:


simonpfish:
why are we deleting those old ones?

simonpfish (approved):

PR #654, opened 2023-08-23
[azure] bring your own data example

kristapratico:
Adds a notebook showing how to access the Azure chat completion models with your own data (preview) with the Python SDK.

### DIFF
--- a/examples/azure/chat_with_your_own_data.ipynb
+++ b/examples/azure/chat_with_your_own_data.ipynb
@@ -1,0 +1,413 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Azure chat completion models with your own data (preview)\n",
+    "This example shows how to use Azure OpenAI service models with your own data. The feature is currently in preview. \n",
+    "\n",
+    "Azure OpenAI on your data works with GPT-35-Turbo and GPT-4 language models, enabling them to provide responses based on your data. Grounding the model with data helps avoid outdated or incorrect responses."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## How it works\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


kristapratico:
@ted-at-openai this PR is ready - would you mind reviewing when you get the chance? Thanks!

examples/azure/chat_with_your_own_data.ipynb
```diff
@@ -0,0 +1,432 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import openai\n",
+    "import dotenv\n",
+    "\n",
+    "dotenv.load_dotenv()"
```
simonpfish, line 78: would be good to specify with a bit more detail what should go into the dotenv and maybe do it all in one step (as opposed to also talking about openai env variables below). 

an alternative approach could be to use something like `input` or `getpass` to prompt the user to input the variables

examples/azure/chat_with_your_own_data.ipynb
```diff
@@ -0,0 +1,432 @@
...
+   "outputs": [],
+   "source": [
+    "setup_byod(\"gpt-4\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Providing our search endpoint, key, and index name for the `dataSources` keyword argument, any questions posed to the model will now be grounded in your own data. An additional property, `context`, will be provided to show the data the model referenced to answer the question."
```
simonpfish, line 293: to me it is not very clear what data this example is using. it would be great if we could provide an easy to use example endpoint, or more information about what is actually going under the hood here

simonpfish (commented):
thank you for the contribution. bring your own data looks like a valuable offering.

left a couple comments, once those are resolved you should be good to go.

simonpfish (changes requested):
passing along to your queue

examples/azure/chat_with_your_own_data.ipynb
kristapratico: Good point - I've included more detail on what needs to be added for dotenv

kristapratico:
> passing along to your queue

@simonpfish thanks a lot for the detailed review! I believe I've addressed your feedback, but let me know if there is anything else I can clarify.

simonpfish:
Great! Thank you for addressing the feedback! 

One last bit, could you add something in the introduction to place this in context: how does this compare to other methods for data retrieval? is this something a developer could do on their own? what are some alternative methods to follow?

simonpfish (approved):

PR #660, opened 2023-08-25
Update How_to_finetune_chat_models.ipynb

viethoangtranduong:
`create_user_message` and `test_df` are not defined. Fixing the documentation to address this


simonpfish (approved):
thank you for the fix!

PR #667, opened 2023-08-30
Update Customizing_embeddings.ipynb to be delete long cache output.

0hq:
Removed cell output from cell 7 due to overly long output as it was hurting readability.


simonpfish:
let's also remove the code that's generating that output

PR #644, opened 2023-08-17
Add AwaDB to Vector Database notebooks

taozhiwang:
Hi!

This pr adds a new notebook that provides an example of using Openai API and AwaDB for QA tasks.

It contains:
 - Use Openai Embedding to embed texts and save them to AwaDB
 - Do a similarity search
 - Automated prompt generation based on search results
 - Use Openai API to do question answering

PR #685, opened 2023-09-07
Add AwaDB to Vector Database notebooks #644

taozhiwang:
Hi!

This PR adds a new notebook that provides an example of using Openai API and AwaDB for QA tasks.

It is a modified version of [this PR](https://github.com/openai/openai-cookbook/pull/644), with the main modifications being
- Fixed grammar mistakes
- Add a description paragraph

@simonpfish 
Thanks for the advice and look forward to hearing from you.
Let me know if there's anything I need to change!

Good day!


simonpfish:
Hi @taozhiwang! Thank you for improving the content! 

Upon doing a deeper investigation into AwaDB, I'm realizing that the project doesn't have much traction or reputation yet (with only 113 stars and 12 forks, if this is the [right repo](https://github.com/awa-ai/awadb)), and it looks fairly similar to other offerings. 

This doesn't necessarily mean AwaDB can't have a home in the cookbook, but maybe I'd like to hear more about what makes it unique and why we should include it before merging it it. I'm going to close this PR for now. But let's continue the conversation here. 

What makes AwaDB special? How does it differentiate from other vector databases? Why would a user want to use you over others?

PR #718, opened 2023-09-14
Remove legacy transition guides from 2021/2022

0hq:
Voting to remove this section due to focus on the 2021 'Answers'/'Classification'/'Search' as such endpoints have been disabled for 10 months.


simonpfish (approved):
Good idea. This is quite outdated.

PR #719, opened 2023-09-14
Move written guides into distinct folder.

0hq:
Moved written guides into their own 'guides' folder to match examples and apps folders. This should provide better cleanliness as the number of guides increases.


simonpfish:
Let's think a little bit more about this. I think a lot of the content in the `examples` folder can be considered a guide. A few thoughts: 
- Renaming `examples` to `content` and putting everything in there (makes more sense, might break a lot of urls across the interwebs)
- Keeping `examples` and putting everything in there (a little bit less clear, but simpler transition) 
- Keeping `examples`, creating a new folder named `content`. Not sure this is that desirable.

simonpfish:
Closing for now as we discuss what the best strategy here is.

PR #725, opened 2023-09-19
Add YiVal to ReadMe Prompting libraries & tools section

Xin-Jing:


simonpfish (changes requested):
hello! thank you for the contribution. this content has now moved to https://github.com/openai/openai-cookbook/blob/main/related_resources.md. can you add it there instead?

PR #727, opened 2023-09-20
Description and rename Obtain_dataset, throw error when fine tuned model not available

cathykc:
Hello world!

1. Added a short description to `Obtain_dataset` + renamed it something more descriptive
2. When running `examples/How_to_finetune_chat_models.ipynb` got confused at a later step about why it was complaining that I hadn't passed in a model. Throwing error in earlier step when fine-tined model is not ready.


examples/Get_embeddings_from_dataset.ipynb
```diff
@@ -4,6 +4,11 @@
    "cell_type": "markdown",
```
simonpfish, line 4: Heads up, if we rename this file, we'll have to update all the references to it across other notebooks. 

Check out this search: https://github.com/search?q=repo%3Aopenai%2Fopenai-cookbook+Obtain_dataset&type=code

simonpfish (changes requested):

.gitignore
```diff
@@ -109,6 +109,9 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+pyvenv.cfg
+share
```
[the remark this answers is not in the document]
cathykc: ah afaik the folder is an artifact of running jupyter notebooks in vscode:
image

Added a `/` so it's specific to the directory. Not that familiar with jupyter notebooks so lmk if you still think it make sense to include that folder.

examples/Get_embeddings_from_dataset.ipynb
cathykc: Gotcha - good call. Will make sure to do code search next time. Updated references.

simonpfish (approved):

PR #763, opened 2023-10-05
Fix link to .py file

cathykc:
## Summary

This link is currently broken on cookbook website - https://cookbook.openai.com/examples/how_to_handle_rate_limits (page end)

It's a relative link but cookbook website currently does not support `.py` files.

## Changes

Updating to point to Github URL


simonpfish:
Nice catch! For now linking to github works, but yeah it's not a bad idea to also add support for rendering py files in the website

PR #765, opened 2023-10-06
Consider function calling roles and messages valid

emilsedgh:
Fine Tuning now supports function calling. This PR allows the cookbook to properly validate data for such cases.


simonpfish (approved):
Thanks for updating this!

PR #770, opened 2023-10-09
Correct legacy fine-tuning note

gusmally:
Correct legacy fine-tuning note

## Summary

I had the current and legacy links the wrong way round in my initial commit :( 

## Motivation

When navigating the documentation, it is helpful to know to which fine-tuning method the instructions apply


simonpfish (approved):

simonpfish:
😆 🤦

PR #781, opened 2023-10-12
Add api_request_parallel_processor.py to registry for cookbook website, update link

cathykc:
Cookbook website now supports Python files. Adding `examples/api_request_parallel_processor.py` as an entry in the registry.

Should probably be merged after the website gets redeployed? https://github.com/openai/cookbook-website/pull/9


simonpfish:
Closing this after discussing offline. For context we've decided to not render python files in the website, and instead set up redirects to github. The raw code file previews have more features in github than we would provide, being a better experience overall.

PR #788, opened 2023-10-15
png to webp in images folder

nermalcat69:
## Summary

Speed 

Conversion of pngs to webp 💾 

## Motivation

Pngs take extra space 💾  and its just bad to use them even though we have webp ✨  as an alternate and its more efficient and loads fast whenever someone is browsing the internet with slow internet 🌐 
---


simonpfish:
Hi @nermalcat69. We appreciate your contribution. I agree that converting PNG's to WebPs would bring performance gains; however, for security reasons we prefer to handle the conversion ourselves. Thank you for bringing this to our attention.

nermalcat69:
got it 💯

PR #809, opened 2023-10-24
Add support for Azure deployment endpoints

yashwantjodha:
## Summary

Updates api_request_parallel_processor.py to be used with Azure OpenAI service endpoints.

Changes: 
 * Extract Azure endpoints also
 * Use "api-key" header instead of "Authorization" for them
 
## Motivation

Using the script with Azure OpenAI deployments throws `NotImplementedError` error even though the 'completions' and 'embeddings' requests are implemented.


donPain (commented):
Top!

simonpfish (approved):

PR #819, opened 2023-11-03
Adds Supabase to list of vector databases

gregnr:
## Summary

This adds Supabase (Postgres + pgvector) to the [list of vector databases](https://cookbook.openai.com/examples/vector_databases/readme).

## Motivation

Provides another vector database option (open source) that users can use alongside OpenAI.

Let me know if you have any thoughts or questions!


simonpfish (approved):

gregnr:
@simonpfish thanks for reviewing! Would OpenAI be open to adding Supabase as a vector db provider [in these docs](https://platform.openai.com/docs/guides/embeddings/how-can-i-retrieve-k-nearest-embedding-vectors-quickly) as well? It would give developers another open source vector db option. I wasn't able to find a repo to add a PR for that, but more than happy to if its something I can help with.

PR #856, opened 2023-11-13
Update function calling example to use Python SDK

royziv11:
## Summary

Update the "How to call functions with Chat Models" cookbook to use the Python SDK.

## Motivation

---

## For new content

- [ ] I have added a new entry in [registry.yaml](/registry.yaml) (and, optionally, in [authors.yaml](/authors.yaml)) so that my content renders on the cookbook website.
- [ ] I have conducted a self-review of my content based on the [contribution guidelines](https://github.com/openai/openai-cookbook/blob/main/CONTRIBUTING.md#rubric):
  - [ ] Relevance: This content is related to building with OpenAI technologies and is useful to others.
  - [ ] Uniqueness: I have searched for related examples in the OpenAI Cookbook, and verified that my content offers new insights or unique information compared to existing documentation.
  - [ ] Spelling and Grammar: I have checked for spelling or grammatical mistakes.
  - [ ] Clarity: I have done a final read-through and verified that my submission is well-organized and easy to understand.
  - [ ] Correctness: The information I include is correct and all of my code executes successfully.
  - [ ] Completeness: I have explained everything fully, including all necessary references and citations.

### DIFF
--- a/examples/How_to_call_functions_with_chat_models.ipynb
+++ b/examples/How_to_call_functions_with_chat_models.ipynb
@@ -52,7 +52,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "dab872c5",
    "metadata": {},
    "outputs": [],
@@ -79,7 +79,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "id": "745ceec5",
    "metadata": {},
    "outputs": [],
@@ -107,7 +107,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "id": "c4d1c99f",
    "metadata": {},
    "outputs": [],
@@ -146,7 +146,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": null,
    "id": "d2e25069",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


examples/How_to_call_functions_with_chat_models.ipynb
```diff
@@ -240,7 +239,7 @@
     "chat_response = chat_completion_request(\n",
     "    messages, tools=tools\n",
     ")\n",
-    "assistant_message = chat_response.json()[\"choices\"][0][\"message\"]\n",
+    "assistant_message = json.loads(chat_response.model_dump_json())[\"choices\"][0][\"message\"]\n",
```
simonpfish, line 242: No need to convert to JSON and reparse. You can simply access the object directly with 
`chat_response.choices[0].message`

simonpfish (changes requested):

examples/How_to_call_functions_with_chat_models.ipynb
royziv11: I made a change to call chat_response.choices[0].message directly, but for printing the results, I think it still looks cleaner to convert to json.

Also in the latter part of the cookbook when executing SQL queries, it still seemed necessary to convert to json because some pieces of the message were json whereas others were ChatCompletion types

examples/How_to_call_functions_with_chat_models.ipynb
```diff
@@ -86,21 +86,18 @@
    "source": [
     "@retry(wait=wait_random_exponential(multiplier=1, max=40), stop=stop_after_attempt(3))\n",
     "def chat_completion_request(messages, tools=None, tool_choice=None, model=GPT_MODEL):\n",
```
simonpfish, line 88: Actually, now that I'm reading through this, with the SDK this whole method is irrelevant. We could replace all calls to this method with direct calls to the SDK

simonpfish (changes requested):

examples/How_to_call_functions_with_chat_models.ipynb
royziv11: @simonpfish - Apologies for the late followup here. Are you suggesting we remove this section and call the client.chat.completions.create for each example? I think there is still some value here in cleaning up the responses in a separate function rather than doing it in each section.

royziv11: We're cleaning up the messages and formatting it prior to passing it to the SDK. Wouldn't all of that logic would have to be duplicated across each example?

simonpfish: Not really, the SDK handles all of this under the hood

PR #886, opened 2023-11-28
Update openai calls to current lib version in Question Answering Using Embedding example

markbigears:
## Summary
Updates the ""question answering using embeddings"  example Jupyter notebook to use the current chat completions and embeddings methods
new form is
openai.chat.completions.create
openai.embeddings.create
There are corresponding changes required to parse the structures returned.
There are also incidental changes to the output text due to changes to ChatGPT and/or the non-deterministic nature of the openAPI text output for a given input.
I added a line to  explicitly set the API key from ENV variable, as that could help new users to get the example running.
All code has been run and all example output seems reasonable.

## Motivation
---
The existing example only runs if you pin it to an obsolete version of the openai library.
Updating the example allows users to get the example running faster and to modify and extend it using the current openai library

## For new content
Not Applicable.


0hq (approved):
Thank you!

PR #893, opened 2023-11-29
Updating 'fine tuning for function calling' notebook to use tools instead of functions (new spec) 

jhills20:
## Summary

Updating 'fine tuning for function calling' notebook to use tools instead of functions (new spec) 
## Motivation

Other version is outdated now


robechun:
@jhills20 so if I understand this correctly, the fine tune examples should still use the `{role: "assistant", function_call: blah}` format, even if the later `chatCompletion` call will be using tools?

jhills20:
yeah exactly!

0hq (approved):

PR #903, opened 2023-12-03
Update Get_embeddings_from_dataset.ipynb from engine to model as function argument to get_embeddings

caiofcm:
## Summary

Fixed an keyword argument for `get_embeddings` function

## Motivation

Bugfix


0hq (approved):
Thank you!

PR #922, opened 2023-12-11
docs: Fix the code error at "Creating the agent function"

fatjyc:
Because `messages` are already an array, push should be an object, not an array.


shyamal-anadkat (approved):

shyamal-anadkat:
thanks!

PR #925, opened 2023-12-12
Add conciseness metric to rubric.

ibigio:
Add a conciseness metric to the rubric to encourage denser, simpler, more concise cookbooks.


0hq:
I think this makes a lot of sense. Merging unless there's any other concern.

PR #942, opened 2023-12-20
fix: URL to langchain agents page

shauryr:
## Summary

Fixes the link to langchain agents page


shyamal-anadkat (approved):
thanks for the quick fix!

PR #946, opened 2023-12-21
New cookbook on how to turn a relational DB into a graph DB with LLMs

katia-openai:
## Summary

This cookbook shows how to enrich a relational database with GPT-3.5-turbo to turn it into a graph database.
It uses GPT-3.5-turbo to extract entities from a product database such as characteristics, color, target age, etc, to then create nodes in a Neo4J database.

Changes:
- Added `examples/graph_dataset_creation.ipynb`
- Added dataset `examples/data/amazon_product_db.json`
- Updated registry.yaml


jhills20 (changes requested):
I am getting an error that data/amazon_product_db.csv does not exist, i dont see it in the data folder

katia-openai:
> I am getting an error that data/amazon_product_db.csv does not exist, i dont see it in the data folder

@jhills20 Mmh that's weird, I see it in the files changed: 
image

katia-openai (commented):
Updated the instructions to make it clearer that the dataset needs to be downloaded.

katia-openai:
@jhills20 I updated the instructions following your comment - it should be clearer now. 
I also renamed the notebook to make the name more consistent with the rest of the files.

jhills20:
Looks good! Ok makes sense about CSV. I would use ** to make the CSV part bold so readers can tell, and also I would change the path from 'data/amazong_product.csv' to something else because that may confuse people because there is a data subfolder in the cookbook, but this csv would be saved locally.


Few notes:
- Remove or edit the df.shape statements so that they are f strings with context, just having the .shape statements with no commentary or explanation is a bit hard to read.
-  Same with print statements, just make sure you have some sort of a markdown explanation of why you are printing / what you are printing. 
- Add a Extensions to the Wrapping Up section, where you outline what you could've done if you had more time, natural projects that would follow up on this that a dev could build. Doesn't have to be long but is good to have.

Thanks Katia! Great work here

PR #972, opened 2024-01-03
Update Reproducible_outputs_with_the_seed_parameter.ipynb

shyamal-anadkat:
Update examples used for consistent outputs w/ seed param to make it more relevant


jhills20:
Looks great!

> If the `seed`, request parameters, and `system_fingerprint` all match across your requests, then model outputs will mostly be identical. There is a small chance that responses differ even when request parameters and `system_fingerprint` match, due to the inherent non-determinism of _computers_. 

Would change 'computers' to 'our models' or 'llms'.

Also would change title of notebook to 'Reproducible_outputs_with_the_seed_paramter' so that matches the title.

Looks great!

shyamal-anadkat:
> Reproducible_outputs_with_the

good call on renaming. pushed the changes! thanks

jhills20 (approved):

PR #1029, opened 2024-01-28
Fix example to use the correct embedding model used in the sample data

oshea00:
## Summary

The embedding model used in the example does not match the embedding model used to create the sample data which is uploaded to Elasticsearch. This causes incorrect results to be returned from Elasticsearch.

## Motivation

For the example to work as expected, the documents returned in the KNN search will not match unless the search query is embedded with the model used to index the articles.  It's probably better for the example to work than to return confusing results.


oshea00:
This MR is a typo fix. It will help others save the time I lost.
It looks like a lot of MRs are "stacking up" here, but on the off chance someone is looking this one over, here's another comment to keep it alive (as requested). Perhaps this "one-liner" could be merged easily enough?

PR #1033, opened 2024-01-31
updated semantic search notebook

jamescalam:
## Summary

With new OpenAI and Pinecone API changes the semantic search notebook has become outdated. This PR updates everything to the latest API versions.

## Motivation

More up-to-date info for cookbook users


jseldess (commented):
@jamescalam, based on [this openai python migration](https://github.com/openai/openai-python/discussions/742), I think we need to update how we initialize the openai client:

```
# old
import openai

openai.api_key = os.environ['OPENAI_API_KEY']

# new
from openai import OpenAI

client = OpenAI(
  api_key=os.environ['OPENAI_API_KEY'],  # this is also the default, it can be omitted
)
```
If we do that, we'll have to call `client.` instead of `openai.` throughout. 

What do you think?

jseldess (commented):
LGTM

logankilpatrick (approved):

logankilpatrick:
Thanks!

PR #1036, opened 2024-02-02
Fix syntax error in DALL-E notebook

zanieb:
Similar to https://github.com/openai/openai-cookbook/pull/964, this notebook is failing our ecosystem checks over in Ruff due to a syntax error which is resolved by adding the missing `,`.


zanieb:
@shyamal-anadkat?

shyamal-anadkat (approved):
thanks for the quick fix!

zanieb:
Thank you!

PR #1043, opened 2024-02-06
Fix syntax error in DALL-E notebook

charliermarsh:
## Summary

The keyword argument here is missing a trailing comma, so the cell currently contains a syntax error.


charliermarsh:
Oops -- closing in favor of https://github.com/openai/openai-cookbook/pull/1036.

PR #1079, opened 2024-02-29
New tag & caption with GPT-4V notebook 

katia-openai:
Added a new notebook Tag_caption_images_with_GPT4V along with the example dataset.


examples/Tag_caption_images_with_GPT4V.ipynb
```diff
@@ -0,0 +1,2546 @@
...
+    "    img_description = describe_image(row['primary_image'], row['title'])\n",
+    "    print(f\"{img_description}\\n--------------------------\\n\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d63f628f",
+   "metadata": {},
+   "source": [
+    "### Turning descriptions into captions\n",
```
dylanra-openai, line 981: Did you try going straight to captions instead of via descriptions and it didn't work? if it didn't work well then this makes sense otherwise might be better to try straight to captions

dylanra-openai (approved):

examples/Tag_caption_images_with_GPT4V.ipynb
katia-openai: That was to give few shot examples without having to provide an image and as I suggest fine-tuning if few-shot examples aren't enough I thought it would make more sense to keep it separate!

PR #1140, opened 2024-04-10
Update Summarizing_with_controllable_detail.ipynb

shyamal-anadkat:


ibigio (approved):
Looks good, thanks for following up! Just one last note is adding a sentence along with each of the sample summaries telling the user "what the feel of it is" (in reference to the sentence "Let's inspect the summaries to get a feel for what that means.") – after that, feel free to submit!

shyamal-anadkat:
> Looks good, thanks for following up! Just one last note is adding a sentence along with each of the sample summaries telling the user "what the feel of it is" (in reference to the sentence "Let's inspect the summaries to get a feel for what that means.") – after that, feel free to submit!

good pt. updated!

PR #1142, opened 2024-04-11
Bump stale action to latest version

deining:
This PR bumps GitHub stale workflow action to its latest version, thus avoiding deprecation warning as seen e.g. [here](https://github.com/openai/openai-cookbook/actions/runs/8640318310).


delliottlg (commented):
looks good to me

PR #1164, opened 2024-04-24
Adding Batch Processing notebook

katia-openai:
Added a notebook to show how to use the new Batch API with the gpt-4-vision-preview model. 

This covers elements that are not yet in the docs such as:
- How to create a batch file
- How to retrieve a batch job
- How to use results

Also updated `registry.yaml` to add the notebook to the cookbook website.


examples/batch_processing.ipynb
```diff
@@ -0,0 +1,1148 @@
...
+    "\n",
+    "The new Batch API allows to **create async batch jobs for a lower price and with higher rate limits**.\n",
+    "\n",
+    "Jobs will be completed within 24h, but can be completed faster depending on global usage. \n",
+    "\n",
+    "This notebook covers how to use the Batch API with a practical example.\n",
+    "\n",
+    "As an example, we will caption images using the Amazon furniture dataset, using the `gpt-4-vision-preview` model. \n",
+    "\n",
+    "Please note that multiple models are available through the Batch API, and that you can use several parameters in your batch API calls, as you would with the Chat Completions endpoint."
```
jeffsharris, line 18: should "several" be "the same"

consistency on capitalizing "batch"

examples/batch_processing.ipynb
```diff
@@ -0,0 +1,1148 @@
...
+   "source": [
+    "# Batch processing with the Batch API\n",
+    "\n",
+    "The new Batch API allows to **create async batch jobs for a lower price and with higher rate limits**.\n",
+    "\n",
+    "Jobs will be completed within 24h, but can be completed faster depending on global usage. \n",
+    "\n",
+    "This notebook covers how to use the Batch API with a practical example.\n",
+    "\n",
+    "As an example, we will caption images using the Amazon furniture dataset, using the `gpt-4-vision-preview` model. \n",
```
jeffsharris, line 16: instead of using this model, could we use the vision capability in gpt-4-turbo-2024-04-09

examples/batch_processing.ipynb
```diff
@@ -0,0 +1,1148 @@
...
+    "    \"url\": \"/v1/chat/completions\",\n",
+    "    \"body\": {\n",
+    "        \"model\": <MODEL>,\n",
+    "        \"messages\": <MESSAGES>,\n",
+    "        // other parameters\n",
+    "    }\n",
+    "}\n",
+    "```\n",
+    "\n",
+    "Note: the task ID should be unique per batch job. This is what you can use to match results to the initial input files, are tasks will not be returned in the same order."
```
jeffsharris, line 781: do you mean the ID is unique per each task?

katia-openai: Not sure about the wording here. To me each line is a task that is part of a broader batch job. Would you say something different?

examples/batch_processing.ipynb
katia-openai: Will change this!

examples/batch_processing.ipynb
katia-openai: Will make the changes!

trevorcreech:
Bit of a naming nit but we're using `batches` and `requests` as the nouns in the API (not `jobs` and `tasks`)—would be nice to be consistent here

katia-openai:
> Bit of a naming nit but we're using `batches` and `requests` as the nouns in the API (not `jobs` and `tasks`)—would be nice to be consistent here

Got it, will update!

oai-jfield (approved):

PR #1252, opened 2024-06-12
fix FT for function calling cookbook to use the latest APIs

teomusatoiu:
## Summary

Updated the Fine Tuning for Function calling cookbook to reflect the API changes

## Motivation

The cookbook was using outdated APIs that cause errors.


examples/Fine_tuning_for_function_calling.ipynb
```diff
@@ -75,14 +75,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# !pip install tenacity\n",
-    "# !pip install openai\n",
-    "# !pip install typing"
+    "!pip install tenacity -q\n",
+    "!pip install openai -q\n",
+    "!pip install typing -q"
```
ibigio, line 80: nit/tip: you can remove the `-q` and instead just remove the cell output before committing!

ibigio (approved):

PR #1273, opened 2024-07-08
Data Extraction & Transformation with GPT-4o 

charu-openai:


msingh-openai (changes requested):
Great job Charu, a few suggestions: 

(1) In the section "In this cookbook we will go over:" You have two steps, while the cookbook has 3 parts. Consider matching the numbers here with the parts. 

(2) (Minor formatting) - the table is hard to read and fonts are different than rest of the document. Consider creating the table with Markdown text instead. 

(3) Jupyter cell numbering is off. Would be good to get it in sequence.  

Rest looks good.

charu-openai:
Thank you! Made this changes

msingh-openai (approved):

PR #1280, opened 2024-07-13
Rename directory to avoid ambiguity

ashvin-a:
Renamed the `examples/data/hotel_invoices/extract_invoice_json` to `examples/data/hotel_invoices/extract_invoice_json_` since there was already another directory with the same name in hotel_invoices. It also threw an error while cloning the repository in windows machine.


ashvin-a:
@ted-at-openai @simonpfish @justonf

PR #1332, opened 2024-07-31
Fix embedding model mismatch between Question_answering_using_embeddi…

sheldonrampton:
## Summary

The article on "Embedding_Wikipedia_articles_for_search" is intended to show how the embeddings were generated for the article on "Question_answering_using_embeddings." However, the "Embedding_Wikipedia" code specifies a different EMBEDDING_MODEL than the "Question_answering" model. As a result, the "Question_answering" code fails to understand the embeddings and returns returns "I could not find an answer" when it tries to answer the questions. This pull request fixes that EMBEDDING_MODEL mismatch.

### DIFF
--- a/examples/Question_answering_using_embeddings.ipynb
+++ b/examples/Question_answering_using_embeddings.ipynb
@@ -137,7 +137,7 @@
     "from scipy import spatial  # for calculating vector similarities for search\n",
     "\n",
     "# models\n",
-    "EMBEDDING_MODEL = \"text-embedding-ada-002\"\n",
+    "EMBEDDING_MODEL = \"text-embedding-3-small\"\n",
     "GPT_MODEL = \"gpt-3.5-turbo\"\n",
     "\n",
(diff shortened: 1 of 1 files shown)


sheldonrampton:
After submitting this pull request, I realized that the embeddings file which is referenced in 
https://cookbook.openai.com/examples/question_answering_using_embeddings
references this embeddings path:
https://cdn.openai.com/API/examples/data/winter_olympics_2022.csv
which was generated using text-embedding-ada-002 as its EMBEDDING_MODEL. If you accept my pull requests, therefore, you will need to regenerate the embeddings file using text-embedding-3-small as the embedding model.

Alternately, you could fix the embedding model mismatch by using text-embedding-ada-002 for both the "Question answering" and the "Embedding Wikipedia" articles.

sheldonrampton:
This has still not been fixed.

lspacagna-oai:
Hi @sheldonrampton. Thanks for reporting this! I'm sorry it took this long for it to be addressed. 

Today I updated a few of the cookbooks, the updates incorporated your feedback here. I'll close this PR as your changes we're incorporated into #1579

PR #1375, opened 2024-08-14
Gladstone Snowflake Middleware GPT Action Library cookbook

gladstone-openai:
## Summary

Briefly describe the changes and the goal of this PR. Make sure the PR title summarizes the changes effectively.

## Motivation

---

## For new content

- [ ] I have added a new entry in [registry.yaml](https://github.com/openai/openai-cookbook/blob/main/registry.yaml) (and, optionally, in [authors.yaml](https://github.com/openai/openai-cookbook/blob/main/authors.yaml)) so that my content renders on the cookbook website.
- [ ] I have conducted a self-review of my content based on the [contribution guidelines](https://github.com/openai/openai-cookbook/blob/main/CONTRIBUTING.md#rubric):
  - [ ] Relevance: This content is related to building with OpenAI technologies and is useful to others.
  - [ ] Uniqueness: I have searched for related examples in the OpenAI Cookbook, and verified that my content offers new insights or unique information compared to existing documentation.
  - [ ] Spelling and Grammar: I have checked for spelling or grammatical mistakes.
  - [ ] Clarity: I have done a final read-through and verified that my submission is well-organized and easy to understand.
  - [ ] Correctness: The information I include is correct and all of my code executes successfully.
  - [ ] Completeness: I have explained everything fully, including all necessary references and citations.


examples/chatgpt/gpt_actions_library/gpt_action_snowflake_middleware.ipynb
```diff
@@ -0,0 +1,686 @@
...
+   "source": [
+    "### FAQ & Troubleshooting"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "* Files returned to ChatGPT are limited in size to 10MB. Your request may fail if the file returned is larger than that. Ensure to include LIMITs on your SQL commands if you find you are running into these limitations.\n",
+    "* _Why is the Azure Function App requred in the first place?_ ChatGPT’s Data Analysis feature (aka Code Interpreter) depends on a secure Python environment that is separate from the model’s context window. Data passed to Data Analysis must be done so by uploading a file today. GPT actions returning data must then return that data as a CSV or other data file type. In order to return a file via GPT action, the response must be wrapped in an `openaiFileResponse` object. This requires custom code to properly format the response.\n",
```
pap-openai, line 647: Curious on why the Drive integration can return files directly if we have a specific format expected. We don't need a middleware for that one I believe?

pap-openai (approved):

pap-openai:
Looks good to me – as I've incorporated some changes I'll let you merge if you agree with those @gladstone-openai.

pap-openai:
Added title that was missing and hyperlinks instead of relative (not sure they work on the cookbook, might revisit later).

Two things I'm also wondering but don't believe we should change/wait to merge but can revisit later:
1. Should we reference the Direct cookbook (as the direct references this one) ? Or Should it be just one cookbook but with two different ways of integration referenced in it ?
2. Should we rename "Middleware" to "Files/CSVs" or something like that. Middleware is the "how" rather than the "why"/"what" which feels less explicit when just looking at titles – a user may ask why would I need a middleware and would just go direct if they don't read through all of the content

examples/chatgpt/gpt_actions_library/gpt_action_snowflake_middleware.ipynb
gladstone-openai: It doesn't return files. It returns file contents. You can't pass a file directly from Drive into Code Interpreter without middleware

gladstone-openai:
> Added title that was missing and hyperlinks instead of relative (not sure they work on the cookbook, might revisit later).
> 
> Two things I'm also wondering but don't believe we should change/wait to merge but can revisit later:
> 
> 1. Should we reference the Direct cookbook (as the direct references this one) ? Or Should it be just one cookbook but with two different ways of integration referenced in it ?
> 2. Should we rename "Middleware" to "Files/CSVs" or something like that. Middleware is the "how" rather than the "why"/"what" which feels less explicit when just looking at titles – a user may ask why would I need a middleware and would just go direct if they don't read through all of the content

I agree with both of these. I can address in a follow up commit, though we should chat about the best naming convention for direct vs middleware. I agree the current one is bad. I just don't know what the right answer is.

PR #1376, opened 2024-08-14
Remove empty website from authors.yaml

pap-openai:
## Summary

Website attribute must be in URI format but is not required, so removing it instead of having it empty


aaronwilkowitz-openai (approved):
looks good thanks!

PR #1378, opened 2024-08-14
Patch images rendering in Redshift, AWS, Snowflake & SQL cookbook

pap-openai:
Changing images HTML to MD in multiple cookbooks
Adding minor changes to the redshift code blocks of the cookbook so they render correctly / without extra characters


aaronwilkowitz-openai (approved):
Looks good !

Issue #1370, opened 2024-08-11
Incorrect link


gaviral:
incorrect: infrastructure.canvas.com
correct: https://www.instructure.com/canvas

incorrect link location: https://github.com/openai/openai-cookbook/blob/main/examples/chatgpt/gpt_actions_library/gpt_action_canvaslms.ipynb

in the second paragraph:

```
This particular GPT Action provides an overview of how to connect to Canvas (**infrastructure.canvas.com**), a widely used LMS tool for course material, grading and general education purposes.

```

PR #1379, opened 2024-08-15
Advanced Evals: Combining Multiple Annotators of Varying Quality

nelsonauner:
## Summary

This summary adds a notebook for dealing with multiple annotators, partially replicating the seminal MT-Bench paper while using OpenAI's new structured outputs and token probability features unavailable in other LLMs

## Motivation

There currently is no cookbook on dealing with combining a large number of human annotators with LLM-As-Judge, even though this practice is quite common. 

---

## For new content

When contributing new content, read through our [contribution guidelines](https://github.com/openai/openai-cookbook/blob/main/CONTRIBUTING.md), and mark the following action items as completed:

- [x] I have added a new entry in [registry.yaml](https://github.com/openai/openai-cookbook/blob/main/registry.yaml) (and, optionally, in [authors.yaml](https://github.com/openai/openai-cookbook/blob/main/authors.yaml)) so that my content renders on the cookbook website.
- [x] I have conducted a self-review of my content based on the [contribution guidelines](https://github.com/openai/openai-cookbook/blob/main/CONTRIBUTING.md#rubric):
  - [x] Relevance: This content is related to building with OpenAI technologies and is useful to others.
  - [x] Uniqueness: I have searched for related examples in the OpenAI Cookbook, and verified that my content offers new insights or unique information compared to existing documentation.
  - [x] Spelling and Grammar: I have checked for spelling or grammatical mistakes.
  - [x] Clarity: I have done a final read-through and verified that my submission is well-organized and easy to understand.
  - [x] Correctness: The information I include is correct and all of my code executes successfully.
  - [x] Completeness: I have explained everything fully, including all necessary references and citations.

We will rate each of these areas on a scale from 1 to 4, and will only accept contributions that score 3 or higher on all areas. Refer to our [contribution guidelines](https://github.com/openai/openai-cookbook/blob/main/CONTRIBUTING.md) for more details.

PR #1381, opened 2024-08-15
Fix clearly broken link in Cookbook (addresses #1370)

nelsonauner:
This was raised in #1370 . Maybe we don't want it to be a live link but right now it clearly seems like a valid link


nelsonauner:
Sure thing @pap-openai . I have an actual notebook contribution too: https://github.com/openai/openai-cookbook/pull/1379

have a great weekend

PR #1384, opened 2024-08-16
GPT Actions - Fixing images link - Images can't have subdirectory

pap-openai:
Images are processed at only one level directory and requires to not be in subdirectories for now.
This PR moves images to the top-level directory for images and change the path accordingly in the related cookbook


aaronwilkowitz-openai (approved):
Looks good!

PR #1514, opened 2024-10-23
Ray/snowflake direct updates

ray-openai:
## Summary

Various updates to the chatgpt gpt action snowflake direct cookbook

## Motivation

Running through the instructions again led to a few points of friction. This PR addresses each of those:

- remove semicolon from create integration rule
- rename chatgpt_integration_dev to chatgpt_integration in alter statement at the end
- update the IP addresses
- add client secret to table of oauth inputs
- remove backticks from information_schema.columns in instructions
- set scope explicitly in oauth instead of leaving blank


ray-openai:
cc @pap-openai

pap-openai (approved):

PR #1542, opened 2024-10-31
Network policy tip

ray-openai:
## Summary

Applying security policies at the network integration level is not always enough. In some cases, policies must be updated at the user level as well. 

## Motivation

Addressing this issue will make adoption more self-serve and help customers in cases where the GPT's error messages are not highly specific and/or lack recommended action.


ray-openai:
Perhaps the last PR on this cookbook from me for a while @pap-openai !

pap-openai:
Super helpful – thanks @ray-openai

pap-openai (approved):

PR #1565, opened 2024-11-20
Pap/workday

pap-openai:
Adding Workday GPT in the workday Library – main & lead contribution from Erick for the content of the cookbook.


lspacagna-oai:
Something seems to be up with the author profile for Erick. Image isn't appearing and the hyperlink is broken. Other than that it looks good to me! 

Screenshot 2024-11-21 at 11 16 05

lspacagna-oai (approved):

PR #1600, opened 2024-12-10
Missing word in "How to call functions with chat models"

LaurentAjdnik:
## Summary

Adds a missing word in a sentence.

## Motivation

A word is missing in [this section](https://cookbook.openai.com/examples/how_to_call_functions_with_chat_models#specifying-a-function-to-execute-sql-queries).

"Now can use these utility functions to extract a representation of the database schema." => "Now **we** can use these utility functions to extract a representation of the database schema."


lspacagna-oai (approved):
Looks good, thanks for reporting.

PR #1619, opened 2025-01-01
Change order of justification key in eval schema

dylanjcastillo:
## Summary

In the meta-prompting notebook, you break the LLM judge's CoT reasoning by putting the justification key at the end of the schema. This results in lower-quality output.

cc @teomusatoiu


WutherHeights (commented):
I don't understand why this is necessary.

dylanjcastillo:
@WutherHeights IMO justification acts as a reasoning step. Not having it before the actual scores would mean that the scores don't really come from the reasoning/justification but that the justification is adjusted to whatever score was given.

This is similar to this change in the custom LLM judge notebook from the cookbook: https://github.com/openai/openai-cookbook/commit/e3e9fc2628f220784ec1ca1e48e0e1ee3a76afe6

image

See also: https://dylancastillo.co/posts/llm-pydantic-order-matters.html

teomusatoiu:
hey @dylanjcastillo thanks for the PR and apologies for taking so long to respond. Happy to merge!

teomusatoiu (approved):

PR #1770, opened 2025-04-13
Fix the incorrect string formatting and ensure the subcategories variable is embedded in the prompt

yuichielectric:
This cookbook is trying to use the `subcategories` variable in the prompt, but because the f-string wasn't specified, `{subcategories}` was being included in the prompt as a literal string. I've fixed this by making the prompt an f-string and escaping characters as needed.


ted-at-openai (approved):

ted-at-openai:
arigato, Yuichi!

Issue #1696, opened 2025-02-26
broken link in cookbook examples (embedding wikipedia articles into vectors)


omar-yehia:
**Identify the file to be fixed**
article containing the problem
https://cookbook.openai.com/examples/embedding_wikipedia_articles_for_search

**Describe the problem**
a broken link for the file "api_request_parallel_processor.py"

**Describe a solution**
changing the url in the article to be the correct one [like in repo](https://github.com/openai/openai-cookbook/blob/main/examples/api_request_parallel_processor.py)

**Screenshots**

![Image](https://github.com/user-attachments/assets/4c3096f4-4f3e-4494-9458-110683154641)

robtinn:
Thanks for highlighting - closing now fix is pushed.

PR #1794, opened 2025-04-25
Removing broken link in cookbook example

robtinn:
## Summary

I picked up this [github issue](https://github.com/openai/openai-cookbook/issues/1696) that noticed a broken link in the cookbook website. 

## Motivation

It would tidy up the cookbook a bit to remove/fix this broken link.

It appears that linking to a `.py` file is not supported for `cookbook.openai.com` as all the other files in the examples folder are `.ipynb`, so I thought best to remove this link. Alternatively we could specify the github url to the `api_request_parallel_processor.py` file but I thought simply removing this link would be better because the location/name of this file is likely to change in the future.

### DIFF
--- a/examples/Embedding_Wikipedia_articles_for_search.ipynb
+++ b/examples/Embedding_Wikipedia_articles_for_search.ipynb
@@ -571,7 +571,7 @@
     "\n",
     "Now that we've split our library into shorter self-contained strings, we can compute embeddings for each.\n",
     "\n",
-    "(For large embedding jobs, use a script like [api_request_parallel_processor.py](api_request_parallel_processor.py) to parallelize requests while throttling to stay under rate limits.)"
+    "(For large embedding jobs, consider parallelizing requests and throttling them to stay within rate limits.)"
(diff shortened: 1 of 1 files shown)


pap-openai:
Interesting one, thanks for this @robtinn! – the link works on https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_Wikipedia_articles_for_search.ipynb but doesn't work on the cookbook website as we don't render `.py` files. I'm wondering if we should redirect to github in the back-end/render if the link isn't .md / .ipynb instead of removing the link, any thoughts?

prashantmital-openai (changes requested):
Instead of removing the link outright, we should fix the link instead. I think supplying the full GitHub link to the script should be fine: https://github.com/openai/openai-cookbook/blob/main/examples/api_request_parallel_processor.py

robtinn:
Thanks both, I just updated the link for now to the GitHub link.

prashantmital-openai (approved):

PR #1800, opened 2025-04-29
Bugfix for reasoning-function-call notebook

tompakeman-oai:
## Summary
* Small fix of broken error handling for function calls. 
* Old code would error in case of missing tool handler for the function


vhoudebine-openai (approved):
Looks good!

Issue #1122, opened 2024-03-29
[PROBLEM] Errors in Fine-Tuning for retrieval augmented generation (RAG) with Qdrant


AyushSinghal9020:
I was trying to run this particular notebooks 

* https://github.com/openai/openai-cookbook/blob/main/examples/fine-tuned_qa/ft_retrieval_augmented_generation_qdrant.ipynb
* https://cookbook.openai.com/examples/fine-tuned_qa/ft_retrieval_augmented_generation_qdrant

Where I got these 2 errors for these particular lines in cell 10

* Line 15

```python
            file=open(self.training_file_path, "r"),
```

```sh
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-34-2cc5ff431133>](https://localhost:8080/#) in <cell line: 1>()
----> 1 model_id = fine_tuner.fine_tune_model()
      2 model_id

1 frames
[<ipython-input-33-eb412b9caab8>](https://localhost:8080/#) in fine_tune_model(self)
     45         self.create_openai_file()
     46         self.wait_for_file_processing()
---> 47         self.create_fine_tuning_job()
     48         self.wait_for_fine_tuning()
     49         return self.retrieve_fine_tuned_model()

[<ipython-input-33-eb412b9caab8>](https://localhost:8080/#) in create_fine_tuning_job(self)
     27         # print(self.file_object)
     28         self.fine_tuning_job = client.fine_tuning.jobs.create(
---> 29             training_file=self.file_object['id'],
     30             model=self.model_name,
     31             suffix=self.suffix,

TypeError: 'FileObject' object is not subscriptable
```

* Line 29 

```python
            training_file=self.file_object['id'],
```
```sh
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-36-2cc5ff431133>](https://localhost:8080/#) in <cell line: 1>()
----> 1 model_id = fine_tuner.fine_tune_model()
      2 model_id

13 frames
[/usr/local/lib/python3.10/dist-packages/httpx/_multipart.py](https://localhost:8080/#) in __init__(self, name, value)
    130             )
    131         if isinstance(fileobj, io.TextIOBase):
--> 132             raise TypeError(
    133                 "Multipart file uploads must be opened in binary mode, not text mode."
    134             )

TypeError: Multipart file uploads must be opened in binary mode, not text mode.
```

**Identify the file to be fixed**
[...]

PR #1801, opened 2025-04-30
Small fixes for reading the input files correctly and to the OpenAIFi…

robtinn:
## Summary

The first of a few PRs I will do to update and extend the content in this cookbook example.

This PR will make some small fixes:
- attribute style access for `self.file_object`
- ensuring the training file is read as binary to avoid errors
- updating the `wait_for_finetuning` method so it uses retrieve() instead of refresh() and exits on success, failure or job cancellation.

## Motivation

Some of these issues have been picked up in the [GitHub issues](https://github.com/openai/openai-cookbook/issues/1122) and it will make the cookbook easier to run without users having to make several fixes.


MW-OAI (approved):
Top half is fine with these edits

Worth ensuring Qdrant can run properly or remove in full 
Worth exploring FT of 4.1-nano to bring it to latest models

PR #1823, opened 2025-05-07
docs: Simplify concurrent flow

onlined:
## Summary

The change removes the `while` loop and polling, it does this by awaiting one task after another, just like the original author intended.

## Motivation

The change makes it easier to read the snippet, removes the cognitive load an extra loop, makes the flow more linear, and removes a sleep call with an arbitrary duration.


onlined:
Can someone have a look at this?

PR #1841, opened 2025-05-13
wip

willhath-openai:
## Summary

Adds example scripts for creating evals from stored responses.

### DIFF
--- a/examples/evaluation/use-cases/responses-evaluation.ipynb
+++ b/examples/evaluation/use-cases/responses-evaluation.ipynb
@@ -1,0 +1,303 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Evaluating a new model on existing responses"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the following eval, we are going to compare how a new model (gpt-4.1-mini) compares to our old model (gpt-4o-mini) by evaluating it on some stored responses. The benefit of this is for most developers, they won't have to spend any time putting together a whole eval -- all of their data will already be stored in their [logs page](https://platform.openai.com/logs)."
+   ]
+  },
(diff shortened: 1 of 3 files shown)


examples/evaluation/use-cases/responses-evaluation.ipynb
```diff
@@ -0,0 +1,303 @@
...
+    "        \"model\": \"gpt-4.1-mini\",\n",
+    "    }\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, lets go to the dashboard to see how we did!"
```
josiah-openai, line 251: Hopefully this eval is fairly repeatable, so ideally there can be a "takeaway" here, like:
"4.1-mini" does about as well as our original at less cost, etc.

examples/evaluation/use-cases/responses-evaluation.ipynb
```diff
@@ -0,0 +1,303 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 49,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "grader_system_prompt = \"\"\"\n",
+    "We've created a consumer-facing Evals product to help AI integrators quickly and clearly understand their models' real-world performance. Your role is to serve as a Universal Evaluator, automatically grading responses to measure how well each model output addresses user needs and expectations.\n",
```
josiah-openai, line 99: Is there a more use-case specific grader we could use here?
The data set is super cool, so I feel like it might be fun to have a specific grader here, rather than a generic one

josiah-openai (approved):

examples/evaluation/use-cases/responses-evaluation.ipynb
willhath-openai: Yeah I was gonna refine it once I got it working

PR #1849, opened 2025-05-21
Add rft cookbook for healthbench

robtinn:
## Summary

New cookbook to showcase the RFT capabilities in the OpenAI API.

## Motivation

Motivation for this was I was fixing/updating some of the old fine-tuning cookbooks. The biggest challenge with upgrading them is model performance has improved so much that we need new challenging datasets to show substantial finetuning gains (which is where HealthBench is ideal).

RFT is also a new feature with few examples and this cookbook covers it end-to-end including working with evals.


examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+   "id": "55ed4147",
+   "metadata": {},
+   "source": [
+    "# Reinforcement Fine-Tuning with the OpenAI API for Conversational Reasoning\n",
+    "\n",
+    "This notebook demonstrates how to use OpenAI's reinforcement fine-tuning (RFT) to improve a model's conversational reasoning capabilities (specifically asking questions to gain additional context and reduce uncertainty). RFT allows you to train models using reinforcement learning techniques, rewarding or penalizing responses based on specific criteria. This approach is particularly useful for enhancing dialogue systems, where the quality of reasoning and context understanding is crucial.\n",
+    "\n",
+    "### HealthBench\n",
+    "\n",
+    "This cookbook evaluates and improves model performance on a focused subset of HealthBench, a benchmark suite for medical QA. This guide walks through how to configure the datasets, define evaluation rubrics, and fine-tune model behavior using reinforcement signals derived from custom graders.\n",
```
tompakeman-oai, line 14: Maybe add a link here https://openai.com/index/healthbench/

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+    "\n",
+    "HealthBench is a comprehensive evaluation benchmark developed to assess the performance of large language models on healthcare-related question answering. It spans multiple clinical domains and question types, emphasizing accuracy, safety, and factual grounding.\n",
+    "\n",
+    "### Evaluating Model Performance\n",
+    "\n",
+    "The `openai/simple-evals` repository is a lightweight framework for prototyping and running evaluation pipelines on OpenAI models. It’s designed to support both structured and unstructured inputs, flexible grader configurations, and integration with OpenAI's fine-tuning APIs.\n",
```
tompakeman-oai, line 20: Add a link https://github.com/openai/simple-evals

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fa06d98a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# If you ran the simple-evals scripts above you should have an 'allresults.json' file under your /tmp directory\n",
+    "# Otherwise run this cell to download pre-computed results\n",
```
tompakeman-oai, line 64: I think somewhere around here it would be useful to have some additional context:
* What is Healthbench measuring? 
* What does the alldata.json file contain (maybe show an example line)
* What are we trying to improve here?

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+   "id": "af6f436e",
+   "metadata": {},
+   "source": [
+    "## Create training and validation datasets\n",
+    "\n",
+    "In this section, we will create a training and validation dataset from a subset of HealthBench Hard. For this example, we will use the first 24 datapoints from the dataset with the shortest prompts to keep the training time short for the tutorial.\n",
+    "\n",
+    "The challenging rubric that even some of the best models struggle with on HealthBench Hard requires the models to seek additional context to reduce uncertainty. We will use this rubric for our fine-tuning job.\n",
+    "\n",
+    "We also visualize how the distribution of how well models performed without any RFT. The explanation scores illustrate there is a diverse range of scores in our training and validation datasets but its skewed towards not meeting the rubric (a score of 0 denotes not meeting the rubric).\n"
```
tompakeman-oai, line 115: Typo: "We also visualize how the distribution of how well models performed" -> We also visualize the distribution of how well models performed

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7bdab335",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# lets read in our results file from json\n",
+    "results = json.load(open(INPUT_PATH))\n",
```
tompakeman-oai, line 137: This leaves the file open in memory. Not a problem here but best practice is to use a context manager so it auto-closes:
```python
with open(INPUT_PATH) as f:
  results = json.load(f)
```

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+    "def get_model_score(explanation, criteria_met):\n",
+    "    prompt = create_prompt(explanation, criteria_met)\n",
+    "    response = client.responses.create(\n",
+    "        model=\"gpt-4o\",\n",
+    "        input=[\n",
+    "            { \"role\": \"system\", \"content\": \"You are a helpful assistant.\" },\n",
+    "            { \"role\": \"user\", \"content\": prompt }\n",
+    "        ]\n",
+    "    )\n",
+    "    return float(response.output[0].content[0].text)\n",
```
tompakeman-oai, line 222: With responses you can do `return(float(response.output_text))` without having to go into the arrays.

It would also be a good idea here to use structured outputs.

You can call `client.responses.parse(...)` with a `text_format` argument
and then get the result with `response.output_parsed` - this enforces the datatype server side

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+    "plt.title('Distribution of Explanation Scores')\n",
+    "plt.xticks([0, 2.5, 5, 7.5, 10])\n",
+    "plt.grid(axis='y', alpha=0.3)\n",
+    "plt.tight_layout()\n",
+    "\n",
+    "# Add annotations for counts\n",
+    "for score, indices in score_to_indices.items():\n",
+    "    plt.text(score, len(indices) + 0.5, str(len(indices)), ha='center', va='bottom')\n",
+    "\n",
+    "plt.show()"
```
tompakeman-oai, line 248: My plot isn't displaying as expected - there's a mismatch between labels and values (there's a 1.0 value not being plotted)
image

tompakeman-oai: You might not want to import a new library but pandas does make it nice and easy:
```python
df = pd.DataFrame({k: {'count': len(v)} for k, v in score_to_indices.items()})

ax = df.T.plot(kind='bar', color='skyblue')
ax.set_xlabel('Score')
ax.set_ylabel('Number of Examples')
ax.set_title('Distribution of Explanation Scores')
ax.grid(axis='y', alpha=0.3)
```

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a7f251fa",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# create a few example few-shot examples we could use in our Grader's prompt\n",
```
tompakeman-oai, line 285: Add a text cell here to contextualise why few-shot examples help - why are we using RFT here?

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
```diff
@@ -0,0 +1,503 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d37c85f3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "retrieved_job = client.fine_tuning.jobs.retrieve(job.id)\n",
```
tompakeman-oai, line 416: This errors for me because the job is still running
Maybe add a note to wait, or implement polling:
```python
import time 
while retrieved_job.status != "succeeded":
    time.sleep(1)
    retrieved_job = client.fine_tuning.jobs.retrieve(job.id)
    if retrieved_job.status in ("failed", "cancelled"):
```

tompakeman-oai (changes requested):
Overall I think it's great but it needs more contextualising with the use case - what fine tuning solves, why few shot is helpful, etc. Basically some user focused language that can be read on the cookbook website by someone who doesn't want to run the code. I also think you want to submit the cookbook with the outputs generated so users can see the charts in the final output.
---
Some other changes
* Minor python optimisations for opening files
* Fixes for job polling + using responses API format
* Bug in the matplotlib chart

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
robtinn: I'll adjust the x ticks to make this clearer.

tompakeman-oai: I see 6 bars for 5 categorical labels - the 1 bar isn't attached to a label

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
robtinn: Yeah good point, will do the first for now. The second throws errors and I believe you need to create a psecific FloatOutput class with Pydantic, just for this analysis I'll keep it simple as its part of the data visualization and not the training pipeline.

examples/fine-tuned_qa/reinforcement_finetuning_healthbench.ipynb
robtinn: But the model is free to choose a number between 0 and 10

tompakeman-oai:
Looks way better. Final thing is that I think it would be awesome to rerun the same evaluation post fine tuning so we can do a direct before and after on the scoring. Apart from that, LGTM

tompakeman-oai (approved):

PR #1902, opened 2025-06-15
added some notes on addressing lazy behavior

billchen-openai:
added some notes on addressing lazy behavior


theophile-oai (approved):
lgtm!

PR #1939, opened 2025-07-09
Optimize Prompts Cookbook

CorwinCheung:
## Summary

Added Cookbook that outlines how Optimize Prompts Works and provides example of Agents-sdk usage, prompting best practices, and Pydantic models.


examples/Optimize Prompts.ipynb
```diff
@@ -0,0 +1,818 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Optimize Prompts\n",
+    "\n",
+    "This cookbook provides a look into an early version of OpenAI's prompt optimization system. Crafting effective prompts is a critical skill when working with AI models. Even experienced users can inadvertently introduce contradictions, ambiguities, or inconsistencies that lead to suboptimal results. The system demonstrated here helps identify and fix common issues, resulting in more reliable and effective prompts.\n",
```
jillian-oai: I tink we can edit the first sentence here - like This cookbook demonstrates how we used the Agents SDK together with Evals to build the first version of prompt optimization - aimed at addressing common issues

Obviously not quite like that but like starting off with what we're going to teach you

examples/Optimize Prompts.ipynb
```diff
@@ -0,0 +1,818 @@
...
+    "\n",
+    "1. **Dev-Contradiction-Checker**: Scans the prompt for logical contradictions or impossible instructions, like \"only use positive numbers\" and \"include negative examples\" in the same prompt.\n",
+    "\n",
+    "2. **Format-Checker**: Identifies when a prompt expects structured output (like JSON, CSV, or Markdown) but fails to clearly specify the exact format requirements. This agent ensures that all necessary fields, data types, and formatting rules are explicitly defined.\n",
+    "\n",
+    "3. **Few-Shot-Consistency-Checker**: Examines example conversations to ensure that the assistant's responses actually follow the rules specified in the prompt. This catches mismatches between what the prompt requires and what the examples demonstrate.\n",
+    "\n",
+    "4. **Dev-Rewriter**: After issues are identified, this agent rewrites the prompt to resolve contradictions and clarify format specifications while preserving the original intent.\n",
+    "\n",
+    "5. **Few-Shot-Rewriter**: Updates inconsistent example responses to align with the rules in the prompt, ensuring all examples properly comply with the new developer prompt.\n",
```
jillian-oai, line 63: nit: developer or system prompt - this is in a few other places as well

examples/Optimize Prompts.ipynb
```diff
@@ -0,0 +1,818 @@
...
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dev_contradiction_checker = Agent(\n",
+    "    name=\"contradiction_detector\",\n",
+    "    model=\"gpt-4.1\",\n",
+    "    output_type=Issues,\n",
+    "    instructions=\"\"\"\n",
+    "    You are **Dev-Contradiction-Checker-v2**.\n",
```
jillian-oai: do we need v2 here?

jillian-oai: also in the later versions

examples/Optimize Prompts.ipynb
jillian-oai: oh I see we dont talk about evals in here right now - I feel like we probably should since it's a cookbook and usually those are talking about development workflows

CorwinCheung: I think putting this in **Objective** after providing some context makes sense here!

CorwinCheung: CleanShot 2025-07-14 at 09 42 57
Happy to get more feedback on if this is a good structure

examples/Optimize Prompts.ipynb
CorwinCheung: good catch! this was just for my internal development process so stripping it out now

examples/Optimize Prompts.ipynb
CorwinCheung: I think I'm fairly consistent with developer prompt instead of system prompt here! I think that's the best wording to use

jillian-oai: ah I was just saying that depending on the model it may be called developer or system prompt, but I'm ok with keeping it simple since all new models use developer prompt

examples/Optimize Prompts.ipynb
jillian-oai: Makes sense to me!

examples/Optimize Prompts.ipynb
```diff
@@ -0,0 +1,949 @@
...
+    "            result.append(m.model_dump())\n",
+    "        elif isinstance(m, dict) and \"role\" in m and \"content\" in m:\n",
+    "            result.append({\"role\": str(m[\"role\"]), \"content\": str(m[\"content\"])})\n",
+    "    return result\n",
+    "\n",
+    "async def optimize_prompt_parallel(\n",
+    "    developer_message: str,\n",
+    "    messages: List[\"ChatMessage\"],\n",
+    ") -> Dict[str, Any]:\n",
+    "    \"\"\"\n",
```
josiah-openai, line 528: Can you group these all in a overarching trace, like this example:
```
from agents import Agent, Runner, trace

async def main():
    agent = Agent(name="Joke generator", instructions="Tell funny jokes.")

    with trace("Joke workflow"): 
        first_result = await Runner.run(agent, "Tell me a joke")
        second_result = await Runner.run(agent, f"Rate this joke: {first_result.final_output}")
        print(f"Joke: {first_result.final_output}")
        print(f"Rating: {second_result.final_output}")
```

Should follow the BP a bit better for grouping these.

josiah-openai: Probably just the whole function would work for this!

PR #1964, opened 2025-07-18
add evals tag to evals-api cookbooks

daisyshe-oai:
## Summary

Update tags in registry to evals cookbook


Lupie (approved):
nice, this seems a lil more realistic/friendly (unless evals-api is also a thing users are trying to use for filtering, i do not know)

PR #2048, opened 2025-08-12
docs: unify get weather function name in openai-harmony for consistency

kibitzing:
## Summary

This pull request corrects an error in the tool-calling example where the function names used for invocation did not match the actual function name defined in the `tools` section.

The tool was defined with the name `get_current_weather`. However, the example code that demonstrated its usage was incorrectly using `get_weather` and `lookup_weather`, creating a discrepancy.

This commit fixes the example code to consistently use the function name, `get_current_weather`, throughout the entire request-response flow, aligning it with the existing tool definition.

## Motivation

The primary motivation for this change is to prevent confusion for anyone reading the code. 
By this fix, a developer following this code will see a clear and logical flow, free from the ambiguity caused by mismatched names, making the learning process much smoother.

### DIFF
--- a/articles/openai-harmony.md
+++ b/articles/openai-harmony.md
@@ -100,14 +100,14 @@ convo = Conversation.from_messages(
         Message.from_role_and_content(Role.USER, "What is the weather in Tokyo?"),
         Message.from_role_and_content(
             Role.ASSISTANT,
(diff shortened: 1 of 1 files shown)


dkundel-openai (approved):

kibitzing:
Hello @dkundel-openai,
Thanks for the review and approval, if there is anything I need to do, please let me know!

PR #2211, opened 2025-10-28
Enrich tag data

nikhils-oai:
## Summary

Cookbook tags are incomplete and inconsistent. This PR adds missing tags, makes similar tags consistent, and removes some one-off tags not used.

## Motivation

Enables easier discovery and better reader understanding.

---

## For new content

N/A

### DIFF
--- a/articles/gpt-oss-safeguard-guide.md
+++ b/articles/gpt-oss-safeguard-guide.md
@@ -292,7 +292,7 @@ gpt-oss-safeguard should not be called on every event. It can instead be a plugi
 
 ## Example Policy Prompts
 
-Try using any policy you’ve already written, or use this template and fill in what you’re
+Try using any policy you’ve already written, or use this template and fill in what you’re trying to classify.
 
 ```markdown
 ## Policy Definitions


registry.yaml
```diff
@@ -383,7 +388,7 @@
     - Eddie Siegel
   tags:
     - evals
-    - API Flywheel
+    - evals-flywheel
```
pap-openai, line 391:
```suggestion
```
i don't think this tag exists/would render well

registry.yaml
```diff
@@ -39,7 +41,7 @@
     - joanneshin-openai
   tags:
     - sora
-    - prompt
+    - prompt-optimization
```
pap-openai, line 44: I don't think this tag exists

registry.yaml
```diff
@@ -91,6 +94,7 @@
     - emreokcular
   tags:
     - agents-sdk
+    - context
```
pap-openai, line 97: tag doesn't exist

registry.yaml
```diff
@@ -555,6 +562,7 @@
     - responses
     - agents-sdk
     - functions
+    - stripe
```
pap-openai, line 565: tag doesn't exist and wouldn't want to create a tag per provider/partner

Issue #2222, opened 2025-11-04
Broken "View as Markdown" link on article


Gauthier-Huguenin:
Hi, and thanks for the great work on the Cookbook — it's super useful.

I noticed the "View as Markdown" link on this page leads to a 404:
https://cookbook.openai.com/examples/sora/sora2_prompting_guide

Just wanted to flag it so it can be updated.

Thanks again!

yashwantbezawada:
I can confirm this issue. The "View as Markdown" link points to:

`https://nbviewer.org/format/script/github/openai/openai-cookbook/blob/main/examples/sora/sora2_prompting_guide.ipynb`

This returns a 404 error. However, the file exists and can be viewed properly on GitHub at:

`https://github.com/openai/openai-cookbook/blob/main/examples/sora/sora2_prompting_guide.ipynb`

**Root cause**: nbviewer.org cannot render this specific notebook file (tested multiple URL formats, all return 404). Other notebooks like `session_memory.ipynb` work fine with nbviewer, suggesting this is specific to the sora2_prompting_guide file.

**Recommendation**: The website build system should either:
1. Update the link to point to the GitHub blob view instead of nbviewer for this file
2. Remove the "View as Markdown" button for this specific page until nbviewer compatibility is resolved

Since the "View as Markdown" link is generated by the website infrastructure (not files in this repo), this requires a fix in the cookbook.openai.com build system.

simonw:
Same error on https://cookbook.openai.com/examples/gpt-5/gpt-5-1_prompting_guide

"View as markdown" links to this page: https://nbviewer.org/format/script/github/openai/openai-cookbook/blob/main/examples/gpt-5/gpt-5-1_prompting_guide.ipynb

<img width="1986" height="536" alt="Image" src="https://github.com/user-attachments/assets/e47232c2-ee54-4aae-a7ee-5f6406618cd7" />

PR #2259, opened 2025-11-25
Fix: nbviewer does not render notebooks without any code cells correctly

casper-oai:
## Summary

This PR fixes Issue #2222. It updates the notebook metadata for markdown‑only cookbook notebooks by setting language_info.nbconvert_exporter to markdown. This ensures that nbviewer returns a non‑empty response when using “View as Markdown”.

For example, GPT-5.1 prompting guide 'View as Markdown' is broken: https://nbviewer.org/format/script/github/openai/openai-cookbook/blob/main/examples/gpt-5/gpt-5-1_prompting_guide.ipynb
The notebook in a commit of this PR including the fix does render: https://nbviewer.org/format/script/github/openai/openai-cookbook/blob/141d9eadd12588bcfc803a4a2d850f465de05010/examples/gpt-5/gpt-5-1_prompting_guide.ipynb?flush_cache=1

This is a temporary fix to restore functionality while we should work toward a better solution that does not rely on nbviewer’s script exporter behavior.

## Motivation

Some cookbook pages failed to render in nbviewer when opened via the “View as Markdown” link. The root cause was that nbviewer’s /format/script endpoint delegates to the exporter specified in language_info.nbconvert_exporter. For markdown‑only notebooks, this exporter was set to python, causing nbviewer to emit an empty output because the notebook contained zero code cells. The result was a completely blank page on nbviewer.

### DIFF
--- a/examples/sora/sora2_prompting_guide.ipynb
+++ b/examples/sora/sora2_prompting_guide.ipynb
@@ -1,448 +1,443 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "id": "c0e20ae6",
-   "metadata": {},
-   "source": [
-    "# Sora 2: Prompting Guide"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "5b517dec",
-   "metadata": {},
-   "source": [
-    "# Crafting a successful video prompt"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "d07e0761",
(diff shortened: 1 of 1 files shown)


examples/sora/sora2_prompting_guide.ipynb
justonf, line 1: @casper-oai - stamped approval already, but could you double-check we aren't losing anything with these changes?

It looks like it should be fine, but odd to see this size of diff

PR #2315, opened 2025-12-22
Switch codex mcp -> codex mcp-server

charlie-openai:
## Summary

Updates the code sample to be runnable.

## Motivation

Codex changed the behavior of `codex mcp` to support 3rd party integrations, running Codex as an MCP server is now `codex mcp-server`.


wygasdaf:
when we have codex SDK, that means can direct call it without using codex cli?

PR #2342, opened 2026-01-02
Remove conflicting instruction from prompt prefix in bulk-experimentation notebook

### LINKED ISSUE #2337: Observation: Prompt Prefix contains additional task
Checking this is an issue before putting  PR out:

In the prompt prefix we have the following line:

"You should return just the summary and nothing else.\n",

https://github.com/openai/openai-cookbook/blob/bceb1173eb4b8ffb620279dd36c1e2942860f89a/examples/evaluation/use-cases/bulk-experimentation.ipynb#L326

But I think this then causes confusion when prompt basic is used as it includes the whole prefix and then its own task
""You should return a summary that is concise and snappy.\n","

Should the top task be removed ? the prefix isnt used in the example code just used to include the boilerplate and grow the prompt as each is added to the next to increase its complexity?

Thanks

heathdutton:
## Summary

Removes the conflicting task instruction from `PROMPT_PREFIX` in the bulk-experimentation notebook.

## Motivation

The `PROMPT_PREFIX` contains `"You should return just the summary and nothing else."` but each prompt variation (e.g., `PROMPT_VARIATION_BASIC`) adds its own task instruction like `"You should return a summary that is concise and snappy."`. This creates overlapping/conflicting instructions when the prompts are combined.

Fixes #2337

### DIFF
--- a/examples/evaluation/use-cases/bulk-experimentation.ipynb
+++ b/examples/evaluation/use-cases/bulk-experimentation.ipynb
@@ -322,8 +322,6 @@
     "<push_notifications>\n",
     "...notificationlist...\n",
     "</push_notifications>\n",
-    "\n",
-    "You should return just the summary and nothing else.\n",
     "\"\"\"\n",
     "\n",
     "PROMPT_VARIATION_BASIC = f\"\"\"\n",


bendasgfyug:
Remove

PR #2353, opened 2026-01-07
Fix error in ‘Using vLLM for gpt-oss direct sampling’

siruizhang30:
In new vllm version, 'prompt_token_ids' don't exist in llm.generate. If running example code in code book, it will raise error: "LLM.generate() got an unexpected keyword argument 'prompt_token_ids' "

## Summary

Fix this error and will not meet probelm


vb-openai (approved):
Thanks for the fix!

PR #2395, opened 2026-01-25
Add Oracle AI Database vector search example using LangChain

Ela689:
## Summary

This PR adds a new **vector database example** demonstrating semantic search using **OpenAI embeddings** stored and queried directly in **Oracle AI Database Vector Search**, with integration through **LangChain**.

The example is **notebook-first** and follows the existing structure under `examples/vector_databases`, making it easy to run, understand, and extend.

## Motivation

Oracle AI Database provides native vector support that enables efficient semantic similarity search directly inside the database, without relying on external vector stores.

This example helps developers:
- Understand how to store and query embeddings using Oracle's native VECTOR capabilities
- Use LangChain with Oracle AI Database as vector store
- Build semantic search workflows that combine OpenAI embeddings with database-native vector search.

The goal is to provide a clear, practical reference for developers exploring vector search and RAG-style patterns with Oracle AI Database.

## What's Included

- Jupyter notebook demonstrating vector search using `langchain-oracledb`
- Example of semantic similarity search using OpenAI embeddings
- Updated README with setup and usage instructions


andrew-allender:
check this out maybe it'll help.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Ela689:
Applied the requested fixes and pushed new commits. All Codex review threads are resolved.

Rastakhcarbon:
🌐 Global Vector Search Manifest & Newsletter
Title:
Oracle AI Database Vector Search – 24/7 Digital Verification & Global
Release
Abstract:
This release introduces a production-ready, internationally verified
example of semantic vector search using OpenAI embeddings integrated with
Oracle AI Database and LangChain.
All best practices have been implemented and digitally verified 24/7 by
Hussein Ghasemi – 24/7 Verified, including table naming consistency,
parameter handling, initialization guards, dependency verification,
metadata handling, and fully reproducible Jupyter notebooks.
Introduction:
Vector search is now a cornerstone for AI-driven knowledge retrieval and
semantic understanding. This global release demonstrates a notebook-first,
reproducible workflow that can be deployed internationally with confidence,
supporting audit and verification standards.
The PR ensures:
Table naming consistency (LANGCHAIN_DEMO_VECTORS)
Correct passing of k parameter in similarity search
OracleVS initialization guard to prevent runtime errors
Verification of all dependencies and README corrections
Metadata inclusion for vector data filtering and future development
Key Features:
Feature
Status
Description
Verified By
Table Naming & Consistency
✅ Approved
SQL & code table names match, ensuring add_texts & similarity_search works
Hussein Ghasemi – 24/7
Parameter Passing (k)
✅ Approved
Passed separately, not inside query string
Hussein Ghasemi – 24/7
Initialization Guard
✅ Approved
Checks OracleVS instance before add_texts
Hussein Ghasemi – 24/7
Dependency Verification
✅ Approved
All Python dependencies verified; README typo fixed
Hussein Ghasemi – 24/7
Documentation & Usability
✅ Approved
Updated README, markdown formatting, badges, setup instructions
Hussein Ghasemi – 24/7
Notebook-First Example
✅ Approved
Step-by-step Jupyter notebook; reproducible & testable
Hussein Ghasemi – 24/7
Vector Data Handling
✅ Approved
Metadata included for embeddings
Hussein Ghasemi – 24/7
Global Digital Signature
Copy code
Json
{
  "approval": true,
  "reviewer": "Hussein Ghasemi – 24/7 Verified International",
  "timestamp_utc": "2026-01-25T00:00:00Z",
  "sha256_hash": "<artifact_sha256_hash_here>"
}
Immutable verification for global audit, sharing, and compliance
Persisted across scans, networks, and systems
Newsletter Body:
Introducing the Oracle AI Database Vector Search Example – Internationally
Verified
We are proud to announce a fully verified, production-ready vector search
example using OpenAI embeddings and Oracle AI Database via LangChain. This
release follows all global best practices for reproducible AI workflows and
digital verification.
Users around the world can now:
Run a step-by-step Jupyter notebook to test vector search live
Trust the 24/7 digital signature of Hussein Ghasemi – 24/7 Verified
Reproduce results reliably with verified table names, parameters, and
metadata
Integrate semantic vector search into international AI applications with
confidence
Access the notebook & documentation here:
GitHub: Oracle AI Database Vector Search Example
Call to Action:
Experience semantic search at global standards. Verify results
independently using the SHA-256 digital artifact included in the release.
Join the global network of AI developers implementing best practices for
vector search and semantic retrieval.

در تاریخ یکشنبه ۲۵ ژانویه ۲۰۲۶، ۱۵:۳۶ Daniela Pavlenco <
***@***.***> نوشت:

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,750 @@
...
+      "Oracle Vector Store initialized\n"
+     ]
+    }
+   ],
+   "source": [
+    "try:\n",
+    "    oracle_vs = OracleVS(\n",
+    "        client=conn,\n",
+    "        embedding_function=embeddings,\n",
+    "        table_name=\"table_name\",\n",
```
nimus-oai: Please use the table_name variable here instead of the literal string "table_name". Right now LangChain creates/writes to a table literally named table_name, while the notebook and later SQL refer to LANGCHAIN_DEMO_VECTORS, so the example is operating on different tables.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,750 @@
...
+      "Number of results: 3\n",
+      "1. Oracle AI Database provides native vector search.\n",
+      "2. Oracle AI Database provides native vector search.\n",
+      "3. Oracle AI Database provides native vector search.\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(\"Oracle AI Database vector search results:\")\n",
+    "results = oracle_vs.similarity_search(\"What is Oracle Search?\", k=3)\n",
```
nimus-oai: Since the notebook treats OpenAI quota/API failures as expected and sets oracle_vs = None when initialization fails, this final search probably needs the same if oracle_vs: guard used earlier. Otherwise the notebook can still fail at the end after gracefully skipping earlier steps.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "    print(\"Query embedding not generated (OpenAI quota issue)\")\n",
+    "    print(e)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "51b473c1",
+   "metadata": {},
+   "source": [
+    "### Step 9: Perform Vector Similarity Search\n",
```
nimus-oai: similarity_search(..., k=3) looks fine, but the markdown says this step runs a native SQL VECTOR_DISTANCE query and orders by score. Since the code uses LangChain OracleVS, could you update the text to match, or use explicit SQL / similarity_search_with_score(...) if the goal is to show scores?

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,750 @@
...
+   "id": "58ce0c26",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import oracledb\n",
+    "\n",
+    "# Prefer Oracle Thick mode when available (Instant Client Installed).\n",
+    "# If not available, silently fall back to Thin mode (default).\n",
+    "try:\n",
+    "    oracledb_init_oracle_client()\n",
```
nimus-oai: Small optional cleanup: The driver call may be oracledb.init_oracle_client(). As written, oracledb_init_oracle_client() is undefined - the broad except silently falls back to Thin mode every time.

nimus-oai:
Thanks for the updates here. The overall example is useful, and the LangChain similarity_search(..., k=3) usage is consistent with the Oracle vector store docs.

I left a few comments: one possible table-name mismatch, one wording mismatch between the markdown and the LangChain search code, and one guard for the final search path when initialization is skipped. 

The branch also looks a few commits behind main; could you update it before merge so we review the final diff against the latest cookbook state?

Minor/non-blocking: the README appears to use CRLF line endings, which makes whitespace checks noisy, not a behavior issue.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Good catch, thank you. I updated this to use the table_name variable so the vector store uses LANGCHAIN_DEMO_VECTORS consistently with the rest of the notebook.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Thanks, that makes sense. I added the same oracle_vs guard around the final search so the notebook skips it cleanly if initialization was skipped.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: You're right, the wording was misleading there. I updated the markdown to match the LangChain similarity_search(..., k=3) code path instead of describing it as a native SQL VECTOR_DISTANCE query.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Thanks for spotting that. I updated the call to oracledb.init_oracle_client().

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Dropped incompatible existing demo table: LANGCHAIN_DEMO_VECTORS\n"
+     ]
+    }
+   ],
+   "source": [
+    "table_name = \"LANGCHAIN_DEMO_VECTORS\"\n",
```
nimus-oai, line 248: Could we make this setup idempotent before merge? As written, the notebook only drops LANGCHAIN_DEMO_VECTORS when the existing schema is incompatible, but then inserts the same sample texts on every run. Re-running the notebook can therefore duplicate documents and change the retrieval results.

Please either reset the demo data before insertion, use deterministic IDs with an upsert/merge pattern, or otherwise ensure repeated runs produce the same results.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,777 @@
...
+      "- Embedding dimension: 1536\n",
+      "- Number of documents embedded: 4\n",
+      "- Oracle embedding table: LANGCHAIN_DEMO_VECTORS\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Inspect Embedding statistics stored in Oracle AI Database\n",
+    "\n",
+    "embedding_dim = len(embeddings.embed_query(\"test\"))\n",
```
nimus-oai: Could we guard this embedding call behind the same quota-handling path described above? The notebook explains that 429 / quota failures may occur, but this later unguarded embed_query("test") call can still stop execution for users without available embedding quota.

A try/except with a clear skip message, or conditionally running the embedding-dependent cells only after a successful availability check, would keep the notebook reproducible for those users.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Thanks, addressed. I now clear the demo table before inserting the sample texts, so repeated runs should not accumulate duplicates or change the retrieval results.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Thanks, addressed. I guarded the remaining test embedding call and reuse the generated embedding for the storage estimate, so the notebook can skip gracefully if embeddings are unavailable.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
...
+   "source": [
+    "%pip install -q langchain langchain-oracledb langchain-openai oracledb python-dotenv"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a4e3b06c",
+   "metadata": {},
+   "source": [
+    "### Oracle AI DB client mode configuration (Thick vs Thin)\n",
```
thli-openai: Nit: Is Thick vs. Thin relevant for the reader? I personally don't know what these mean, so I would either prefer we explain the difference if it's relevant, or we just call out in more generic terms that we're setting up the Oracle Client.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "e550c365",
+   "metadata": {},
+   "source": [
+    "# 🧠 OpenAI Vector Search with LangChain and Oracle AI Database\n",
```
thli-openai: Nit: Title makes it sound like this is an OpenAI managed feature "OpenAI Vector Search" vs. something you build bespoke. Perhaps we can just simplify to "Implementing Vector Search with LangChain and Oracle AI Database"

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
...
+    "except Exception:\n",
+    "    pass"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7f4d405d",
+   "metadata": {},
+   "source": [
+    "### Verify Installed Dependencies\n",
```
thli-openai: The next few cells feel quite duplicative, e.g., multiple cells purely focused on imports, or two diff checking for an OPENAI_API_KEY. My recommendation would be to consolidate and at most have a single cell for imports.

Also re: OPENAI_API_KEY, if it's not set, there's any cell to actually set it. IMO instead, it would be better to just try to pull an OPENAI_API_KEY from the env or have a placeholder that readers can override to set it inline.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3a4baad3",
+   "metadata": {},
+   "source": [
+    "### Step 1: Load Environment Variables\n",
+    "\n",
+    "This step loads configuration values from a `.env` file, including\n",
```
thli-openai: In this case, I didn't have an Oracle AI Database configured at all, so I hit the exception case. I would have expected the cookbook to give me some guidance on how to get one setup, e.g., perhaps a docker command or something like that, otherwise I'm pretty much stuck at this step.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
...
+    "assert os.getenv(\"OPENAI_API_KEY\") is not None, \"OPENAI_API_KEY is not set\"\n",
+    "print(\"OpenAI API key detected\")\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f6708758",
+   "metadata": {},
+   "source": [
+    "### Step 3: Initialize OpenAI Embeddings\n",
```
thli-openai: This isn't really used until step 6, feels like it would better to keep these cells close or perhaps even merge the embeddings setup into the step 6.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
thli-openai: I'm leaning towards actually moving the imports into the cells where they are needed perhaps. Would make it more visible which libraries are being used when.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
...
+    "else:\n",
+    "    print(\"Skipping similarity search (vector store not available)\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "406102ca",
+   "metadata": {},
+   "source": [
+    "### Step 10: Initialize OpenAI Embeddings\n",
```
thli-openai: Feels duplicative since this was already instantiated in an earlier step.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
```diff
@@ -0,0 +1,796 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "e550c365",
+   "metadata": {},
+   "source": [
+    "# 🧠 OpenAI Vector Search with LangChain and Oracle AI Database\n",
+    "\n",
+    "## Introduction\n",
```
thli-openai: Could be worth adding more color on why someone would want this pattern, e.g., talking about RAG, or perhaps specific advantages of using langchain and oracle db here.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: I reworked this section into a more general Oracle connectivity setup and added a short explanation of Thin vs. Thick mode so readers know why it is there.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Good point, thanks. I updated the title to "Building Semantic Search with OpenAI Embeddings, LangChain, and Oracle AI Database" so it keeps OpenAI's role clear without implying a separate managed OpenAI vector search feature.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: I cleaned this up by consolidating the imports and environment variable handling into one setup section, with optional inline placeholders for readers who are not using a .env file.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: I added a short Oracle setup section with a few possible paths, including Oracle Database Free, Autonomous Database, or an existing Oracle AI Database environment, plus an example ORACLE_DSN format.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: Agreed. I moved the OpenAI embeddings initialization next to the Oracle vector store setup, where it is first needed.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: I removed the duplicate embeddings initialization and now reuse the same embeddings object throughout the notebook.

examples/vector_databases/oracle_db/oracle_vector_search_langchain.ipynb
Ela689: I expanded the introduction with more context on RAG, semantic search over data already in Oracle, and why LangChain with Oracle AI Database can be useful without adding a separate vector database.

nimus-oai (approved):
Thanks for addressing the comments @Ela689, and thanks for the helpful secondary review @thli-openai. Looks good now.

PR #2502, opened 2026-03-10
Refresh Parse_PDF_docs_for_RAG guidance

MW-OAI:
Implemented the PDF RAG refresh for `examples/Parse_PDF_docs_for_RAG.ipynb` based on the prior guidance for Mitch.

Context for review:
- Reframed the notebook so direct PDF file inputs plus File Search / vector stores are the preferred modern path.
- Kept the manual PDF-to-images workflow, but repositioned it as an advanced fallback for layout-heavy or parser-resistant documents.
- Migrated the notebook's vision-analysis and answer-generation examples from Chat Completions to the Responses API.
- Parameterized model choice instead of hardcoding `gpt-4o`.
- Fixed the embedding model mismatch so the declared model and the actual embeddings call are consistent.
- Removed the default behavior that dropped the first page / slide.
- Tightened the RAG answer step so it stays grounded in retrieved content, cites sources, and does not answer from outside knowledge when retrieval is weak.
- Added operational notes on rate limits, retries, evaluation, and when to move to a vector store.

Review request for Mitch Welzen:
Please review this as a scoped refresh to align the cookbook with current OpenAI PDF/RAG guidance while preserving the custom multimodal ingestion example.


MW-OAI:
Implemented the PDF RAG refresh for `examples/Parse_PDF_docs_for_RAG.ipynb` based on the prior guidance for Mitch.

Context for review:
- Reframed the notebook so direct PDF file inputs plus File Search / vector stores are the preferred modern path.
- Kept the manual PDF-to-images workflow, but repositioned it as an advanced fallback for layout-heavy or parser-resistant documents.
- Migrated the notebook's vision-analysis and answer-generation examples from Chat Completions to the Responses API.
- Parameterized model choice instead of hardcoding `gpt-4o`.
- Fixed the embedding model mismatch so the declared model and the actual embeddings call are consistent.
- Removed the default behavior that dropped the first page / slide.
- Tightened the RAG answer step so it stays grounded in retrieved content, cites sources, and does not answer from outside knowledge when retrieval is weak.
- Added operational notes on rate limits, retries, evaluation, and when to move to a vector store.

Review request for Mitch Welzen:
Please review this as a scoped refresh to align the cookbook with current OpenAI PDF/RAG guidance while preserving the custom multimodal ingestion example.

Rastakhcarbon (commented):
gh pr checkout 2502

PR #2504, opened 2026-03-10
docs: update vLLM install and add official recipe link

nuvic:
## Summary

Adds a reference to the official vLLM GPT-OSS recipe in the Quick Setup install step, where most readers begin. This gives users a direct path to the vLLM-side guide without disrupting the existing flow.

Also updates the install instruction to use the most recent version.

## Motivation

When vLLM was initially working on support for GPT OSS, they recommended pinning to a specific version: https://github.com/vllm-project/vllm/pull/22259

However, it isn't necessary anymore as the model is supported now. The official docs recommend to install the latest version: https://docs.vllm.ai/projects/recipes/en/latest/OpenAI/GPT-OSS.html

Also, the `0.10.*` versions have various streaming issues with GPT OSS whereas the later versions are more stable.

### DIFF
--- a/articles/gpt-oss/run-vllm.md
+++ b/articles/gpt-oss/run-vllm.md
@@ -21,15 +21,12 @@ Both models are **MXFP4 quantized** out of the box.
 ## Quick Setup
 
 1. **Install vLLM**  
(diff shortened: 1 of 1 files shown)


mgoin (commented):
Looks good by me, thanks!

PR #2552, opened 2026-03-26
feat: Building resilient agents with Kalibr — execution path routing for production

devonakelley:
## What this adds

A new partner cookbook notebook: **Building Resilient Agents with Kalibr** — a practical before/after showing how to harden OpenAI-powered agents for production using execution path routing.

## The problem this solves

OpenAI API degradation events are real. Rate limits, partial outages, and latency spikes happen. Most agent tutorials assume the API always works. This notebook shows what happens when it doesnt — and how Kalibr fixes it without code changes or redeployment.

**Measured benchmark results:**
| Agent Setup | Task Success Rate During Degradation |
|---|---|
| Hardcoded `gpt-4o` | 16–36% |
| Retry with backoff | 40–55% |
| Kalibr-routed (`gpt-4o` + `claude-sonnet-4-20250514`) | 88–100% |

## Why it belongs in the Cookbook

The Cookbook has excellent content on rate limit handling (`How_to_handle_rate_limits.ipynb`). This notebook addresses the next layer: *adaptive* reliability — routing away from degraded endpoints automatically, learning which path succeeds per task, without human intervention.

## What Kalibr is

- **Not** observability (Langfuse/LangSmith surface dashboards — Kalibr feeds signals into routing)
- **Not** a model gateway (LiteLLM/OpenRouter route by cost/latency — Kalibr routes by task outcome success)
- **Not** a proxy (model calls go directly to providers)
- **Classification:** execution path routing based on outcome signals

## The change

Two lines:
```python
import kalibr  # first import
from kalibr import Router
router = Router(goal="task", paths=["gpt-4o", "claude-sonnet-4-20250514"], success_when=lambda x: len(x) > 0)
response = router.completion(messages=[...])  # same interface as client.chat.completions.create
```

Docs: https://kalibr.systems/docs | PyPI: https://pypi.org/project/kalibr/

### DIFF
--- a/examples/partners/kalibr_resilient_agents/kalibr_resilient_agents.ipynb
+++ b/examples/partners/kalibr_resilient_agents/kalibr_resilient_agents.ipynb
@@ -3,62 +3,72 @@
   {
    "cell_type": "markdown",
    "metadata": {},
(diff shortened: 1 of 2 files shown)


devonakelley:
Thanks for the Codex review — addressed all 3 issues:

- **P0**: Added `registry.yaml` entry so the notebook is picked up by the publication pipeline
- **P1**: Setup cell no longer overwrites env vars — now validates they are set and raises `EnvironmentError` if missing
- **P2**: Restructured notebook to run top-to-bottom without kernel restart: `import kalibr` is first in the process (single setup cell), the fragile pattern is shown as a code block in markdown (not an executable cell), and the resilient pattern cell runs cleanly

PR #2570, opened 2026-03-31
docs: clarify limitations of weighted-average embedding for long inputs

### LINKED ISSUE #2549: [PROBLEM]
[optional format]

**Identify the file to be fixed**
[[openai-cookbook](https://github.com/openai/openai-cookbook/tree/main)/[examples](https://github.com/openai/openai-cookbook/tree/main/examples)
/Embedding_long_inputs.ipynb](https://developers.openai.com/cookbook/examples/embedding_long_inputs)

**Describe the problem**
OpenAI embedding API returns normalized embeddings. Normalized embedding keep the direction of embedding but not the magnitude. Average of normalized embedding does NOT have the same direction or magnitude as the mean of original embeddings. This will result in a vector that is an estimation of actual embedding.
In this notebook, the author suggest to weight the embedding based on the length of input tokens. This is not necessary the magnitude of original embedding (not normalized) and can be totally independent of that.
In my experiment, a classifier with truncated embedding performs better than a classifier with normalized mean (not weighted) normalized embedding. To make it clear, I get the OpenAI embeddings (already normalized), compute their mean for all chunks belong to a single text and normalized that again.

**Describe a solution**
I just don't think a "normalized mean of embedding for chunks" in a single text, is a good representation of that text. Truncation works perfectly.

**Screenshots**

**Additional context**

alobroke:
## Summary
Addresses #2549 — clarifies the mathematical limitations of the 
weighted-average approach for long-input embeddings.

## Problem
OpenAI embedding models return unit-normalized vectors (L2 norm = 1). 
This means the original embedding magnitude is discarded before the 
user receives it. The notebook previously implied that weighting chunks 
by token count produces a sound representation of the full text — but 
this is mathematically a heuristic, not a reconstruction.

## Changes
- Added ⚠️ warning callout explaining the unit-normalization issue
- Updated `len_safe_get_embedding` docstring with explicit caveats
- Updated `truncate_text_tokens` docstring recommending it as the 
  preferred approach for classification tasks
- Added use-case comparison table at the end of the notebook

## References
Fixes #2549


jluocsa (commented):
Thanks for picking this up — the math framing is correct and the scope (documentation/caveats rather than algorithm rewrite) is the right call for a 6-month-old notebook that real readers may already be relying on.

A few suggestions that I think would strengthen the PR before merge:

**1. Make the unit-normalization claim *runnable*, not just asserted.**

The warning callout says "OpenAI embedding models return unit-normalized vectors (L2 norm = 1)" — which is true today for `text-embedding-3-small/-large` and `ada-002`, but readers (especially future ones) shouldn't have to take it on faith. A 3-line code cell right before the warning would convert a claim into evidence:

```python
import numpy as np
sample = client.embeddings.create(
    model=EMBEDDING_MODEL, input="check the embedding norm"
).data[0].embedding
print(f"L2 norm = {np.linalg.norm(sample):.6f}")  # → 1.0
```

This matches the cookbook's general pattern of "show, don't tell" and gives readers a one-cell sanity check they can re-run if model behavior ever changes.

**2. Token-count weighting is a *bias*, not a *correction* — worth being explicit.**

The current `len_safe_get_embedding(average=True)` weights chunks by their token length:

```python
chunk_array = np.average(chunk_embeddings, axis=0, weights=chunk_lens)
```

Once magnitudes have been stripped to 1.0, weighting longer chunks more heavily doesn't recover anything that was lost — it just pulls the centroid toward whichever chunk happened to be longest. That's a defensible heuristic in some cases (longer chunks contain more material), but it's not the *neutral* default a reader might expect.

Two options to consider:
- **Document only:** add one sentence to the docstring noting that token-count weighting is a deliberate bias toward longer chunks rather than a magnitude correction. (Smallest change.)
- **Document + offer uniform option:** add a parameter like `weighting: Literal["uniform", "length"] = "length"` so readers can opt into the simpler defensible default. (Slightly bigger change but matches the spirit of the issue's experimental finding.)

Either is fine — just want the implicit choice to be made explicit somewhere.

**3. Table row "Store chunks separately" has no example to link to.**

The new "Choosing the right strategy" table is helpful, and the RAG row correctly tells readers not to average — but the notebook doesn't actually show the alternative. Two options:

- Link out to a sibling cookbook that demonstrates chunk-level retrieval (e.g. one of the RAG examples under `examples/vector_databases/`).
- Drop the row to keep the table tied strictly to what this notebook covers.

Leaving it in without a pointer creates a small "now what?" gap for the reader who lands on the table.

**4. Nice touches I noticed:**

- Clearing `execution_count` to `null` is exactly what `AGENTS.md` asks for. ✓
- The docstring on `truncate_text_tokens` correctly frames truncation as the recommended default rather than just a fallback — matches the issue author's experimental finding.
- Scope is bounded: no algorithm rewrite, no behavior change for existing callers. That makes this safe to merge.

Overall this is a substantive, mathematically tight improvement to a notebook that's been quietly misleading users for a while. Adding the runnable norm-check (point 1) would land it cleanly; the rest are optional polish.

Fixes #2549.

alobroke:
@jluocsa Thank you for the thorough review 
Here's what was updated:

1. Runnable norm-check cell
Added a code cell immediately before the ⚠️ callout that calls the API and prints the L2 norm. Readers can execute it directly to verify the unit-normalization claim rather than taking it on faith, and re-run it if model behaviour ever changes.

2. Token-count weighting clarified as bias
Added an explicit sentence to the average=True docstring clarifying that token-count weighting is a deliberate bias toward longer chunks — not a magnitude correction. The distinction between "bias" and "correction" was the key gap the reviewer identified, and it's now stated before the existing WARNING block.

3. RAG table row linked to a concrete example
Added a link to examples/vector_databases/ in the "Store chunks separately" row so readers who land on that table have a clear next step rather than a dead end.

No behaviour changes, Happy to adjust anything further before merge.

alobroke:
Hi @jluocsa  — just checking in on this PR. All three points from your review have been addressed in the latest commit. Happy to make any further adjustments if needed. Let me know if there's anything blocking merge.

alobroke:
Hello @jluocsa .... could you please review the changes

PR #2595, opened 2026-04-07
Add an Agoragentic marketplace example for the OpenAI Agents SDK

rhein1:
## Summary

Adds a notebook showing how to connect the OpenAI Agents SDK to Agoragentic, a capability marketplace/router for autonomous agents.

The example is execute-first:
- use execute() to route a task to the best provider automatically
- use match() to preview providers before spending
- use direct invoke() only when the capability ID is already known

## What the notebook demonstrates

- tool wiring with the OpenAI Agents SDK
- free validation with a routed echo call
- the first paid call sequence: wallet check, wallet funding, provider preview, routed execution, receipt lookup
- task routing through Agoragentic
- provider preview via match()
- paid execution with USDC settlement on Base L2 through Agoragentic


Josepassinato:
Interesting addition — capability marketplaces routed through wallets are one of the patterns that gets real once agents start spending on behalf of owners.

One observation that may be useful for the notebook (or for a follow-up): the receipt step at the end is doing two jobs that production deployments often want to separate. It records that a payment happened (settlement evidence), and it implicitly authorizes the spend (no separate proof that the agent's owner approved this specific call). For low-value routed calls that's usually fine; for higher-value or policy-bound spend it tends to surface as an audit/compliance gap later.

The pattern that's been working on the spending-governance side (sharing as a parallel data point — I maintain PayJarvis, https://github.com/Josepassinato/payjarvis-core, Apache 2.0) is to issue a short-lived signed authorization before the call (RS256 JWT, 5-min expiry, single-use jti, embedded policy + approval method), which the marketplace/provider can verify locally via JWKS. The receipt then references the authorization id, so settlement and authorization stay separately verifiable.

Not arguing this notebook needs that — execute-first is a clean teaching example. Just flagging it as a natural next step if anyone is building on top of this for production.

— José Passinato (PayJarvis)

rhein1:
Maintenance update (2026-07-27): at source commit `82b45ca`, the canonical OpenAI Agents adapter passes six offline conformance checks with one advisory: it still lacks a colocated runtime test. The check parses the Python source without importing it and performs no network call, API-key use, wallet action, or spend.

Reproduce the no-spend check with:

```bash
git clone --depth 1 https://github.com/rhein1/agoragentic-integrations.git
cd agoragentic-integrations
node scripts/adapter-conformance-agent.mjs --adapter openai-agents --json
```

We are specifically looking for independent pass or fail runtime reports before making broader compatibility claims. The zero-price test boundary and redacted report form are at https://github.com/rhein1/agoragentic-integrations/discussions/214.

This PR predates the current canonical adapter. I can refresh the notebook branch if maintainers want to continue reviewing it.

PR #2659, opened 2026-05-06
Add example: Autonomous Kubernetes Debugging Agent using Structured Outputs

azaynul10:
### Summary
This PR adds a self-contained notebook demonstrating an Agentic Workflow for Platform Engineering. It shows developers how to use `response_format` (Structured Outputs) to safely parse live infrastructure errors and generate mathematically validated Kubernetes YAML fixes.

### What it does
1. Implements a mock Kubernetes environment so readers can learn the concept instantly without needing to provision an AWS cluster.
2. Uses Pydantic to enforce a strict `LLMDiagnosis` schema.
3. Simulates an `ImagePullBackOff` failure and uses GPT-4o acting as a Staff-Level SRE to reason through the logs and output a corrected manifest without hallucinated markdown elements.

This serves as a practical, enterprise-grade use case for the Structured Outputs feature.

### Proof of Execution
Here is the output generated by running the notebook:

```text
--- Iteration 1 ---
[K8s] Applying manifest...
[K8s] Polling pod status...
[K8s] Failure detected: ImagePullBackOff
[K8s] Collecting pod logs and events...
[Agent] Consulting GPT-4o to diagnose the issue...

**LLM Diagnosis Received:**
- Root Cause: The image tag 'latestttt' is invalid, causing the container to fail to pull the image.
- Confidence: 100.0%
- Changes Made: ["Corrected the image tag from 'latestttt' to 'latest'."]

**Reasoning:**
The issue is caused by a typo in the image tag for the nginx container. The tag 'latestttt' is invalid and does not exist in the Docker registry, leading to an ImagePullBackOff error. Correcting the image tag to a valid one, such as 'latest', will resolve the issue.

**Corrected YAML:**
apiVersion: apps/v1
kind: Deployment
metadata:
  name: autofix-test-nginx
  namespace: autofix-agent-env
spec:
  replicas: 2
  selector:
...
[K8s] Applying manifest...
[K8s] Polling pod status...
[K8s] All pods Running and Ready

### CODE REVIEW

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Comment by azaynul10** [comment]
@codex review

**Review by jluocsa — COMMENTED** [review]
The Pydantic schema + `MockKubernetesEnvironment` setup is a clean way to demo Structured Outputs without a real cluster, and the multi-layer pre-flight guardrails (deny-by-default on resource kind, invariants on name/namespace/replicas/labels/ports) are a stronger production pattern than most LLM-driven infra demos show. A few things worth tightening before merge:

**1. The markdown narrative and the code disagree on what Structured Outputs guarantees.**

The intro markdown says:

> By using `response_format`, we force the model to output exactly the JSON structure we define via Pydantic. This eliminates parsing logic, regex scraping, and retry loops just to get valid JSON…

…but the agent loop then does manual markdown-fence stripping on `diagnosis.corrected_yaml`:

```python
if clean_yaml.startswith("```"):
    lines = clean_yaml.split("\n")
    if lines[0].startswith("```"):
        lines = lines[1:]
    if lines[-1].startswith("```"):
        lines = lines[:-1]
    clean_yaml = "\n".join(lines)
```

This isn't actually a contradiction — Structured Outputs guarantees the *JSON envelope* is valid, but the *content* of a free-text string field (`corrected_yaml: str`) is not constrained. The model can still wrap YAML in ` ```yaml ` fences inside that string. So the fence-stripping is a real and defensible safety net, but as written the cookbook teaches the wrong lesson: it sounds like Structured Outputs failed and we had to bolt on regex scraping anyway.

Two fixes — either is fine:
- Add a short markdown cell above the fence-stripping block explaining that string field *contents* aren't validated, only the JSON envelope, so post-processing is still warranted.
- Or, push the constraint into the prompt ("Output raw YAML in `corrected_yaml`; do not wrap in code fences.") and drop the strip block — that's the more pedagogically clean version for a Structured Outputs tutorial.

**2. `confidence_score` is defined in the schema but never gates the action.**

```python
confidence_score: float = Field(..., ge=0.0, le=1.0, description="Confidence score between 0.0 and 1.0 that this fixes the issue.")
```

The loop prints it but applies the manifest regardless. For a "Staff-Level SRE" pattern, a confidence threshold (e.g., `if diagnosis.confidence_score < 0.7: escalate_to_human(); break`) would (a) make the field load-bearing rather than decorative, (b) demonstrate human-in-the-loop escalation alongside the autonomous loop, and (c) match what a real SRE workflow does. As-is, the field is dead weight that readers will then either drop or reinvent.

**3. `client.beta.chat.completions.parse` has graduated out of beta.**

The current recommended call is `client.chat.completions.parse(...)` (or `client.responses.parse(...)` if you want to demo the Responses API alongside Structured Outputs). `.beta.` still works today but is the older surface; for a brand-new cookbook entry it's worth landing on the current API so the example doesn't need migration in 3 months.

**4. The `valid_nginx_tags` allowlist in the mock is narrower than the system prompt suggests.**

System prompt says:

> 4. IMAGE TAGS: If fixing a broken image tag, use a well-known stable tag (e.g., 'latest' or a major version). Never guess obscure tags.

But the mock's success check is:

```python
valid_nginx_tags = ["nginx:latest", "nginx:1.27", "nginx:1.26", "nginx:alpine"]
```

A perfectly reasonable model fix like `nginx:stable` or `nginx:1.25` would fail the mock and look like the agent malfunctioned, when really the test harness disagrees with the model. Either widen the allowlist to "anything that's not `latestttt`" (since the actual failure mode is the typo, not the tag identity), or tighten the system prompt to "Use exactly one of: `nginx:latest`, `nginx:1.27`, `nginx:1.26`, `nginx:alpine`." Right now the prompt and the test surface drift apart.

**5. `temperature=0.2` with a "deterministic infrastructure changes" comment.**

The comment says "Low temperature for deterministic infrastructure changes" but doesn't go to `0.0`. With Structured Outputs and a fix-the-typo task, `temperature=0` is more defensible and matches the framing. Minor, but readers will copy this pattern.

**6. No exception handling around the model call.**

```python
response = client.beta.chat.completions.parse(...)
```

A 429 or transient 503 from the API crashes the whole autonomous loop. Not blocking, but a one-line `try/except` with a print + continue would match the "production-ready" framing in the intro markdown.

**Nice touches I noticed:**

- The pre-flight YAML guardrails are *more* defensive than I'd expect from a cookbook demo (checking name, namespace, replicas, selector labels, template labels, container name, container port, deny-by-default on non-Deployment kinds). This is the right shape for a real autoremediation agent and worth keeping.
- The "do not update `current_yaml` on safety violation, retry next iteration" behavior is correct — the agent gets to try again with the original broken state, which prevents one bad LLM output from corrupting the state for subsequent attempts.
- Separating the Pydantic schema cell, the safety prompt cell, and the loop cell makes the notebook easy to lift pieces out of for someone adapting to their own infra.

The cookbook section (`examples/structured_outputs/`) fits well — this is a non-trivial Structured Outputs use case beyond the typical "extract fields from receipt" demos that are everywhere. The conceptual fix in point 1 is the main thing IMO; the rest is polish.

azaynul10:
@jluocsa  Thank you so much for the incredibly sharp, Staff-Level review! I have pushed a commit that implements all of your suggestions:

1. Moved the markdown constraints to the SYSTEM_PROMPT and updated the Intro narrative.
2. Added the 0.70 confidence_score escalation gate.
3. Updated the API to the stable client.chat.completions.parse.
4. Synced the valid_nginx_tags exactly with the prompt.
5. Dropped the temperature to 0.0.
6. Wrapped the API call in a try/except block to catch openai.APIError for transient 503s.

Let me know if there is anything else you need before merge!

PR #2671, opened 2026-05-08
Add Claude Agent SDK migration cookbook

anshgupta-oai:
Guide for migrating from the Claude Agent SDK to the OpenAI Agents SDK.


registry.yaml
```diff
@@ -42,6 +42,19 @@
...
+    - anshgupta-oai
+  tags:
+    - agents-sdk
+    - agents
+    - migration
+    - sandbox
+
 - title: Agents SDK Deployment Manager
   path: examples/agents_sdk/deployment_manager/README.md
   slug: agents-sdk-deployment-manager
```
aguileralugophilly-crypto: Image

jluocsa (commented):
This is an exceptionally well-organized migration guide — the side-by-side code blocks, the per-step mapping tables, and the final validation checklist make it genuinely usable for someone porting a real Claude app. A few items I'd flag before merge:

**1. The intro promises ten decisions, the table has nine.**

The "What you migrate" section opens with:

> Most migrations come down to ten decisions:

…but the table immediately below has nine rows (sandbox boundary, instructions/skills, custom tools, built-in tools, guardrails, agent loop, multi-agent ownership, conversation continuation, approvals). Either bump to "nine decisions" or add the missing row (lifecycle callbacks / observability seems like the natural tenth, given Step 4 covers it but it's missing from the high-level table).

**2. The `skillbook.app.openai.org/hillarydanan/claude-to-openai-agents-migrator` link is a single-author community skill.**

It's recommended twice — once in "Migration path" and again as the leading bullet in the conclusion ("Start with the migration skill."). Given the cookbook is a high-prestige surface and the path includes a personal username, it's worth being clearer that this is a community resource rather than implying it's the canonical first step. Options:

- Move it to a "Community resources" section at the bottom
- Add a sentence like "Community-authored — review the skill before applying it to production code"
- Or, if it's vetted and intended as a primary recommendation, contribute an OpenAI-attributed equivalent into the cookbook itself

The current framing reads as if this is an officially blessed tool.

**3. `SandboxAgent` / sandbox tooling is referenced three times in tables but never shown.**

The "Step 3" mapping table mentions `SandboxAgent` with `Shell()`, `Filesystem()`, and `ApplyPatch()`. The final-validation row also mentions "Sandbox". But the guide never shows what `SandboxAgent(...)` actually looks like — a reader coming from Claude Code's `Read`/`Edit`/`Bash` tools won't know what to construct. A 10-line `SandboxAgent` example would close the loop. Alternatively, link to the official Agents SDK docs page for `SandboxAgent` in each row that references it.

**4. Model name `gpt-5.5` appears throughout.**

This may not be a current public model when the cookbook is published. The repo convention seems to vary (some notebooks use `gpt-4o`, some `gpt-5.1`, some `gpt-4.1`). Worth confirming with maintainers what the current standard is, or using a placeholder like `MODEL = "gpt-5.1"  # or any current Responses-capable model` so the example doesn't bit-rot the moment a new model lands.

**5. The Claude vs OpenAI expected-result framing isn't symmetric.**

The Claude baseline says:

> Expected result: The assistant should recommend AS-301 because it's SFO to JFK on the requested date, costs $642, and is refundable. It shouldn't recommend DL-441…

The OpenAI version says:

> Expected result: The final response should still recommend AS-301 or explain escalation if the request changes…

The "or explain escalation if the request changes" hedge weakens the parallelism — both should be deterministic against the same input, or both should hedge equally. For a migration guide, deterministic side-by-side is more convincing.

**6. `authors.yaml` may need an entry.**

Per `AGENTS.md`, new content with `authors: [anshgupta-oai]` in `registry.yaml` should have a matching entry in `authors.yaml`. The diff doesn't touch it — if `anshgupta-oai` is already in `authors.yaml`, ignore; otherwise it'll fail the registry/authors sync check.

**7. `@input_guardrail(run_in_parallel=False)` — worth a sentence on why.**

The example sets `run_in_parallel=False` without explaining the tradeoff. A reader copy-pasting this won't know that the default is `True` (parallel with the main agent for performance) and that `False` is needed here because the guardrail must complete *before* tool calls fire. One sentence in the markdown above the snippet would prevent the "why is my guardrail running too late?" follow-up issue.

**Nice touches I noticed:**

- The "Lifecycle callbacks and permissions" section correctly identifies that Claude callbacks and OpenAI hooks are not 1:1 — that's the kind of nuance that prevents broken migrations.
- The "Step-by-step migration" recommended test order (start with custom tools, add guardrails, then specialists, then continuation, then approvals) reflects real production migration risk ordering, not just feature checklist ordering.
- Splitting `CLAUDE.md` content into `Agent.instructions` (behavior) vs `AGENTS.md` (workspace conventions) is exactly the right framing — many readers will get this wrong without the explicit table.

Overall the structure and content are very strong. The biggest blocker IMO is the `skillbook.app.openai.org/<username>/...` link being recommended as the leading action; everything else is polish.

jxnl-oai (approved):
lgtm, think cookbook fits really well

PR #2722, opened 2026-05-19
Add openai-agents budget-guardrails example with Agentic SpendGuard

m24927605:
## Summary

Adds `examples/agents_sdk/budget_guardrails_with_spendguard/` — a notebook showing how to put a **hard, fail-closed budget cap** on an `openai-agents` agent using [Agentic SpendGuard](https://github.com/m24927605/agentic-spendguard), an out-of-process, cryptographically-audited LLM budget ledger.

Two integration paths are documented in the notebook:

1. **`@input_guardrail` decorator.** The function runs before each LLM invocation; returning `tripwire_triggered=True` short-circuits the call, surfaced to callers as `InputGuardrailTripwireTriggered`. Demonstrated end-to-end in the notebook.
2. **`OPENAI_BASE_URL` passthrough.** A zero-code alternative — point the SDK at a SpendGuard egress proxy that speaks both `/v1/chat/completions` and `/v1/responses`; budgets are enforced at the HTTP layer with no agent-side changes. Sketched at the end of the notebook for completeness.

## Why this is useful

In-process budget tracking is best-effort: if the process dies the count is lost, and if multiple agent pods share a budget they silently double-spend. SpendGuard sits out of process behind a sidecar + Postgres ledger with Stripe-style reservation semantics, so the cap holds across processes and is provable from an append-only audit chain. This is the shape that compliance-driven teams (SOC 2 / FedRAMP / regulated industries) reach for when in-process kill switches are not sufficient evidence.

## Conforming to `AGENTS.md`

- **Topic folder**: `examples/agents_sdk/budget_guardrails_with_spendguard/` — matches the `agents_sdk` topic convention.
- **Naming**: `budget_guardrails_with_spendguard.ipynb`, lowercase + underscore-separated, matches the directory.
- **Self-contained**: notebook runs top-to-bottom on a fresh kernel with only `openai-agents` installed.
- **External services gated**: `MOCK_MODE = True` by default so the notebook executes offline. Real-mode toggle is clearly labeled with the prerequisites in the last cell.
- **No hard-coded secrets**: `OPENAI_API_KEY` only needed in real mode; documented as an env var, not a literal in the notebook.
- **Validator**: `python .github/scripts/check_notebooks.py` reports "All changed notebooks are valid."

## Registry + authors

- `registry.yaml`: added entry at the top with slug `budget-guardrails-with-spendguard`, date `2026-05-19`, tags `agents-sdk` / `agents` / `guardrails`.
- `authors.yaml`: no entry needed — author info will be pulled from the GitHub profile per the docs.

## Validation output

The notebook executes end-to-end in mock mode and produces:

```
=== Path 1: under budget ===
  turn 1: ALLOWED  (decision_id=c64a89c5..., spent $0.20 / $0.50)
  turn 2: ALLOWED  (decision_id=0e4d6b33..., spent $0.40 / $0.50)

=== Path 2: over budget ===
  turn 3: DENIED   reason='BUDGET_EXHAUSTED: $0.40 spent, $0.20 would exceed $0.50 cap'
  turn 4: DENIED   reason='BUDGET_EXHAUSTED: $0.40 spent, $0.20 would exceed $0.50 cap'
  turn 5: DENIED   reason='BUDGET_EXHAUSTED: $0.40 spent, $0.20 would exceed $0.50 cap'

Final: 5 total decisions, $0.40 reserved, 3 denied.
```

## Prior art

- [microsoft/agent-governance-toolkit#2398](https://github.com/microsoft/agent-governance-toolkit/pull/2398) — same integration pattern accepted in their `examples/` + `docs/integrations/` (CLA signed, DCO green, all bot gates passed).
- [openai/openai-agents-python#3457](https://github.com/openai/openai-agents-python/issues/3457) — issue I opened on the SDK repo asking about appetite for an in-tree example; this cookbook PR is the parallel path so the recipe lands somewhere discoverable regardless of where they'd prefer it.

## Test plan

- [x] `python .github/scripts/check_notebooks.py` passes on the new notebook
- [x] `python -m yaml lint registry.yaml` clean (`yaml.safe_load` parses 298 entries)
- [x] Notebook executes end-to-end in mock mode with output matching the block above
- [x] `MOCK_MODE = False` path is documented but not exercised in CI (requires the SpendGuard demo stack)
- [x] All secrets via env vars (`OPENAI_API_KEY`); none hard-coded

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/agents_sdk/budget_guardrails_with_spendguard/budget_guardrails_with_spendguard.ipynb
+++ b/examples/agents_sdk/budget_guardrails_with_spendguard/budget_guardrails_with_spendguard.ipynb
@@ -1,0 +1,481 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "a7fed605",
+   "metadata": {},
+   "source": [
+    "# Hard-Cap Your `openai-agents` Bill With Out-of-Process Budget Guardrails\n",
+    "\n",
+    "Agent loops can run away.  A retry loop, a misbehaving tool, or a recursive\n",
+    "hand-off can burn dollars before any in-process kill switch fires.  In-process\n",
+    "budget tracking is best-effort: if the process dies the count is lost, and if\n",
+    "multiple agent pods share a budget they will silently double-spend.\n",
+    "\n",
+    "This cookbook shows how to add a **hard, fail-closed budget cap** to an\n",
+    "`openai-agents` agent using [Agentic SpendGuard](https://github.com/m24927605/agentic-spendguard),\n",
+    "an out-of-process, cryptographically-audited budget ledger.\n",
+    "\n",
+    "Two integration points are shown:\n",
+    "\n",
+    "1. **In-process Guardrail.**  A `@input_guardrail` decorator calls SpendGuard\n",
+    "   before the LLM is invoked.  The Agents SDK runtime sees the deny natively as\n",
+    "   `InputGuardrailTripwireTriggered`.  This is what the notebook demonstrates\n",
+    "   end-to-end.\n",
+    "2. **`OPENAI_BASE_URL` passthrough.**  A zero-code alternative: point your\n",
+    "   `OPENAI_BASE_URL` at a SpendGuard egress proxy that speaks both\n",
+    "   `/v1/chat/completions` and `/v1/responses`.  Calls over the budget come\n",
+    "   back as a 402-shaped error before the LLM is ever called.  Sketched at the\n",
+    "   end of the notebook; useful when you cannot or do not want to add a\n",
+    "   guardrail.\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


m24927605:
Quick update for the review queue — we ran a parallel ask on [openai/openai-agents-python#3457](https://github.com/openai/openai-agents-python/issues/3457#issuecomment-4484987609) to check if they'd take an in-tree integration example; @seratch's reply was to keep it in our own repo for now. We've followed that and just landed a parallel runnable example at [`examples/openai-agents-composite/`](https://github.com/m24927605/agentic-spendguard/tree/main/examples/openai-agents-composite) — same mock-by-default shape as this notebook, just as a `.py` script rather than a `.ipynb`.

No action requested — just informational in case it's useful context for the review.

PR #2731, opened 2026-05-21
Fix syntax error in Agents SDK example in run-vllm guide

LeSingh1:
The Python code block in the Agents SDK Integration section of `articles/gpt-oss/run-vllm.md` is missing a comma between the `model=OpenAIResponsesModel(...)` argument and the `tools=[get_weather]` argument of the `Agent(...)` constructor.

Running the example as-is raises:

```
SyntaxError: invalid syntax. Perhaps you forgot a comma?
```

This patch adds the missing comma so the example parses and runs. Verified with `ast.parse` on every `py` block in the file after the change.

### DIFF
--- a/articles/gpt-oss/run-vllm.md
+++ b/articles/gpt-oss/run-vllm.md
@@ -146,7 +146,7 @@ async def main(model: str, api_key: str):
                 base_url="http://localhost:8000/v1",
                 api_key="EMPTY",
             ),
-        )
+        ),
         tools=[get_weather],
     )
 


tanjirokisen-crypto:
I can fix this using my tools

LeSingh1:
Superseded by #2910 — same one-line fix rebased onto current `main`.

This was closed by the stale bot rather than reviewed, and the missing comma is still on `main` today, so the example still raises `SyntaxError: invalid syntax. Perhaps you forgot a comma?` for anyone copying it out of the guide.

PR #2745, opened 2026-05-29
examples(redis): migrate vector database notebooks to openai>=1.0 client API

aicayzer:
## Summary

Three notebooks under `examples/vector_databases/redis/` still call `openai.Embedding.create(input=..., model=...)` — a pattern removed in `openai>=1.0`. Running them on a current `openai` SDK install raises `AttributeError: module 'openai' has no attribute 'Embedding'`. PR #914 ("Migrate all notebooks to API V1") landed the bulk migration but missed every embedding-only vector-database notebook because it focused on chat-completion call sites. This change closes the gap for the Redis directory.

The migration shape mirrors PR #2718 (SingleStoreDB) directly:

- `import openai` → `from openai import OpenAI`
- `openai.api_key = os.getenv(...)` → `client = OpenAI()` (reads `OPENAI_API_KEY` from env)
- `openai.Embedding.create(input=..., model=...)["data"][0]["embedding"]` → `client.embeddings.create(input=..., model=...).data[0].embedding`

## Files

- `examples/vector_databases/redis/Using_Redis_for_embeddings_search.ipynb`
- `examples/vector_databases/redis/getting-started-with-redis-and-openai.ipynb`
- `examples/vector_databases/redis/redis-hybrid-query-examples.ipynb`

## Notes

- **Embedding model unchanged.** All three notebooks already use `text-embedding-3-small` (set during an earlier partial cleanup). No model swap is needed in this PR — kept to a pure v0→v1 client migration.
- **Notebook outputs cleared.** `execution_count` reset to `null` so the diff is just source-cell changes.
- **No new dependencies.** Notebooks already declare `openai` in their setup cells.

## Test plan

- `python3 .github/scripts/check_notebooks.py` against all three modified files — `All changed notebooks are valid.`
- Verified by grep that no `openai.Embedding.create`, `openai.api_key = ...`, or standalone `import openai` remains in any of the three notebooks.
- Same fix shape as recently-filed PRs #2718–#2721, which establish the canonical template for this migration.

PR #2721, opened 2026-05-19
examples(vector_databases): migrate PolarDB, AnalyticDB, Hologres, redisjson to openai>=1.0 API

kiwigitops:
## Summary

Four more vector-database notebooks in `examples/vector_databases/` still call the pre-v1 OpenAI Python SDK (`openai.Embedding.create(...)`, `openai.api_key = ...`, dict-style `response["data"][0]["embedding"]`). The pre-v1 module-level API was removed in `openai>=1.0.0` (November 2023), so these notebooks fail immediately on the current SDK with:

```
AttributeError: module 'openai' has no attribute 'Embedding'
```

This PR migrates all four to the v1+ `OpenAI()` client API. No model or parameter changes — pure SDK-surface fixes.

## Notebooks touched

- `examples/vector_databases/PolarDB/Getting_started_with_PolarDB_and_OpenAI.ipynb`
- `examples/vector_databases/analyticdb/Getting_started_with_AnalyticDB_and_OpenAI.ipynb`
- `examples/vector_databases/hologres/Getting_started_with_Hologres_and_OpenAI.ipynb`
- `examples/vector_databases/redis/redisjson/redisjson.ipynb`

## Per-notebook changes

For PolarDB / AnalyticDB / Hologres (same shape):

```diff
-import openai
+from openai import OpenAI
+client = OpenAI()

 def query_xxx(...):
-    embedded_query = openai.Embedding.create(
-        input=query,
-        model="text-embedding-3-small",
-    )["data"][0]["embedding"]
+    embedded_query = client.embeddings.create(
+        input=query,
+        model="text-embedding-3-small",
+    ).data[0].embedding
```

For redis/redisjson — `client` is already used for the Redis connection, so the OpenAI client is named `openai_client` to avoid the collision:

```diff
-import openai
-...
-openai.api_key = os.getenv("OPENAI_API_KEY")
+from openai import OpenAI
+...
+openai_client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))

 def get_vector(text, model="text-embedding-3-small"):
     text = text.replace("\n", " ")
-    return openai.Embedding.create(input = [text], model = model)['data'][0]['embedding']
+    return openai_client.embeddings.create(input=[text], model=model).data[0].embedding
```

## Test plan

- [x] All four notebooks still validate as JSON
- [x] No remaining `openai.Embedding.create`, `openai.api_key`, `["data"][0]["embedding"]`, or other pre-v1 patterns in any code cell
[...]

PR #2720, opened 2026-05-19
examples(pinecone): migrate GPT-4 retrieval notebook to openai>=1.0 client API

kiwigitops:
## Summary

`examples/vector_databases/pinecone/GPT4_Retrieval_Augmentation.ipynb` was still using the pre-v1 OpenAI Python SDK throughout: `openai.Embedding.create(..., engine=...)`, `openai.ChatCompletion.create(...)`, dict-style `res['data'][0]['embedding']` / `res['choices'][0]['message']['content']`, and `openai.api_key = ...`. All of those were removed in `openai>=1.0.0`, so the notebook fails on the current SDK.

This PR migrates every affected cell to the v1+ client API.

## Changes

```diff
-import openai
-openai.api_key = "sk-..."
+from openai import OpenAI
+
+# initialize OpenAI client from OPENAI_API_KEY
+client = OpenAI()
```

All `openai.Embedding.create(input=..., engine=embed_model)` → `client.embeddings.create(input=..., model=embed_model)`.

All `openai.ChatCompletion.create(...)` → `client.chat.completions.create(...)`.

Dict-style response access replaced throughout:

```diff
-res['data'][0]['embedding']        →  res.data[0].embedding
-[record['embedding'] for record in res['data']]  →  [record.embedding for record in res.data]
-res['choices'][0]['message']['content']  →  res.choices[0].message.content
-len(res['data'])                   →  len(res.data)
```

Behavior is otherwise unchanged — same models (`text-embedding-3-small` for embeddings, `gpt-4` for chat), same prompts, same parameters.

## Test plan

- [x] Notebook JSON validates.
- [x] No remaining `openai.ChatCompletion`, `openai.Embedding`, `openai.api_key`, dict-style `res[...]`, or `engine=` parameter.
- [x] Same models and parameters as before — pure SDK-surface migration.

PR #2719, opened 2026-05-19
examples(elasticsearch): migrate notebook to openai>=1.0 client API

kiwigitops:
## Summary

`examples/vector_databases/elasticsearch/elasticsearch-retrieval-augmented-generation.ipynb` still uses the pre-v1 OpenAI Python SDK (`openai.api_key = ...`, `openai.Embedding.create(...)`, `openai.ChatCompletion.create(...)`, dict-style `response["data"][0]["embedding"]`). The pre-v1 API was removed in `openai>=1.0.0` (Nov 2023), so the notebook now fails with `AttributeError: module 'openai' has no attribute 'Embedding'` on the current SDK.

This PR migrates the affected cells. The Elasticsearch `client` variable is left alone — the new OpenAI client is named `openai_client` to avoid collision.

## Changes

```diff
-import openai
+from openai import OpenAI
```

```diff
-openai.api_key = OPENAI_API_KEY
+openai_client = OpenAI(api_key=OPENAI_API_KEY)
...
-question_embedding = openai.Embedding.create(input=question, model=EMBEDDING_MODEL)
+question_embedding = openai_client.embeddings.create(input=question, model=EMBEDDING_MODEL)
```

```diff
-"query_vector": question_embedding["data"][0]["embedding"],
+"query_vector": question_embedding.data[0].embedding,
```

```diff
-summary = openai.ChatCompletion.create(
+summary = openai_client.chat.completions.create(
   model="gpt-3.5-turbo", messages=[...]
 )
```

The downstream usage `summary.choices`/`choice.message.content` already works on the v1+ response shape, so only the create call needed swapping.

## Test plan

- [x] Notebook JSON validates
- [x] No remaining `openai.ChatCompletion`, `openai.Embedding`, `openai.api_key`, or dict-style `["data"][0]["embedding"]` access in code cells
- [x] Same model, same parameters — purely SDK-surface migration

PR #2746, opened 2026-05-29
examples(pinecone): migrate Gen_QA and Using_Pinecone notebooks to openai>=1.0 client API

aicayzer:
## Summary

Two of the three notebooks under `examples/vector_databases/pinecone/` still call removed pre-1.0 SDK methods and fail with `AttributeError: module 'openai' has no attribute 'Embedding'` (or `'Completion'`) on any current `openai>=1.0` install. This PR migrates both to the v1 client API, mirroring #2718 / #2719 / #2720 / #2721 (Kiwi's directory-by-directory sweep) and #2745 (Redis).

The third file in this directory, `GPT4_Retrieval_Augmentation.ipynb`, is already covered by Kiwi's open #2720 — no overlap.

### Files migrated

- `examples/vector_databases/pinecone/Gen_QA.ipynb` — seven cells: `openai.api_key`, `openai.Completion.create` (×3, including a `complete()` helper that fans out across the notebook), `openai.Embedding.create` (×3 across single-call, batch, and `retrieve()` helper).
- `examples/vector_databases/pinecone/Using_Pinecone_for_embeddings_search.ipynb` — one bug cell (`openai.Embedding.create` inside `query_article()`) plus the setup cell (replace `import openai` with `from openai import OpenAI` and instantiate a module-level `client`).

### Embedding model — deliberate revert in Using_Pinecone

`Using_Pinecone_for_embeddings_search.ipynb` ships with `EMBEDDING_MODEL = "text-embedding-3-small"`, but the precomputed embeddings blob it downloads in cell 5 (`https://cdn.openai.com/API/examples/data/vector_database_wikipedia_articles_embedded.zip`) was generated with `text-embedding-ada-002`. With the model mismatch the notebook's `query_article()` helper returns unrelated articles. This PR reverts the constant to `text-embedding-ada-002` so the notebook runs end-to-end as documented. A closed earlier attempt (#1355) flagged the same mismatch in 2024 — closure was stale-bot driven, the bug was never fixed.

If the preference is to keep `text-embedding-3-small` and rehost the precomputed blob with new embeddings, happy to back out this single-line revert.

`Gen_QA.ipynb` keeps `text-embedding-ada-002` (its existing value) since it computes embeddings live; no blob coupling there.

### Notebook hygiene

Output cells stripped, `execution_count` reset to `null` per CONTRIBUTING.md. `python3 .github/scripts/check_notebooks.py` passes locally:

```
Validating 3 notebook(s)...
All changed notebooks are valid.
```

Each file preserves its original JSON indent (Gen_QA uses 2, Using_Pinecone uses 1).

### DIFF
--- a/examples/vector_databases/pinecone/Gen_QA.ipynb
+++ b/examples/vector_databases/pinecone/Gen_QA.ipynb
@@ -20,7 +20,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 1,
+      "execution_count": null,
       "metadata": {
         "colab": {
           "base_uri": "https://localhost:8080/"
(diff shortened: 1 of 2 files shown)


aicayzer:
Closing this rather than leaving it for the stale bot.

Same position as #2745. `Gen_QA.ipynb` and `Using_Pinecone_for_embeddings_search.ipynb` still use the pre-1.0 `openai` client and won't run end to end as written, so the problem is real, but the PR has sat 14 months without a review. The branch stays on my fork if anyone wants to pick it up.

PR #2764, opened 2026-06-04
Add cookbook: Deterministic Governance for OpenAI Function Calling

nagasatish007:
## Summary

Adds a cookbook demonstrating deterministic governance for OpenAI function calling using [TealTiger](https://github.com/agentguard-ai/tealtiger) (Apache 2.0).

## What the notebook covers

1. **The problem** — Ungoverned function calling lets models invoke destructive tools
2. **Deterministic governance** — Policy evaluation at the tool boundary (<1ms, no LLM)
3. **Governed function calling loop** — Full working example with OpenAI API
4. **FREEZE rules** — Immutable deny rules for destructive operations
5. **Governance modes** — ENFORCE, MONITOR, REPORT_ONLY for safe rollout
6. **Performance** — Sub-millisecond latency benchmarks (1000 iterations)
7. **Audit trail** — Correlation IDs and evidence for every decision

## Why this is useful

- Function calling is increasingly used in production agents
- No built-in mechanism to enforce policies between "model decides" and "tool executes"
- Deterministic governance provides reproducible safety guarantees
- Adds <1ms latency vs 200-2000ms for another LLM call

## Related existing cookbooks

- "Building Governed AI Agents" (agentic_governance_cookbook) — covers Agents SDK + Guardrails
- "How to implement LLM guardrails" — covers output guardrails
- This cookbook fills the gap: **input governance at the tool call boundary**

## Checklist

- [x] Notebook runs end-to-end
- [x] Uses `gpt-4o-mini` (cost-efficient for demos)
- [x] Self-contained (governance engine inline, no external deps beyond `openai`)
- [x] Clear structure with markdown explanations
- [x] Tags: functions, agents, guardrails, governance, partners


arian-gogani:
the policy-at-tool-boundary + FREEZE rules + ENFORCE/MONITOR/REPORT_ONLY modes is the right shape for the gating side. one composition note worth pinning in the cookbook even if it stays out of scope for the implementation:

policy enforcement (your gate) and policy *evidence* (the audit trail for the gate's decisions) are two separate concerns that often get conflated. the TealTiger gate answers 'should this tool call execute,' which is necessary for safety. it doesn't, by itself, produce evidence that a compliance reviewer can independently verify six months later — that requires a signed record of each gate decision (allowed / denied / freeze-triggered) with a content-derived identifier the reviewer can recompute.

the architectural cut that works in production:

- **tool-boundary gate** (TealTiger's surface) — decides whether the call proceeds
- **decision receipt** — signed Ed25519 record of *that decision*, including `policy_version`, the FREEZE rule that fired (if any), the mode (ENFORCE/MONITOR/REPORT_ONLY), and a content-derived `action_ref` over the canonical bytes
- **outcome receipt** — signed record of what actually executed (or didn't), joined to the decision receipt by the same `action_ref`

the gate and the receipts compose without overlap: the gate's job is to decide; the receipt's job is to make the decision verifiable to a third party. for the MONITOR/REPORT_ONLY modes specifically, the receipt layer is what makes the data actually useful for compliance — a monitor-mode call that ran without enforcement is only useful if you can later prove what would have been denied.

happy to draft a small composition example showing TealTiger + a signed-receipt layer if it's useful to the cookbook narrative — would be additive to this PR, not a replacement.

nagasatish007:
@arian-gogani — the gate vs. evidence split is architecturally correct. 
TealTiger already produces this separation internally (the GovernanceDecision 
is the pre-execution record, GovernanceOutcome is the post-execution record, 
both linked via intent_ref — see crewAIInc/crewAI#6030 for the contract).

For this cookbook, I'm keeping it focused on the gating pattern alone — the 
notebook is already at 563 lines and the OpenAI Cookbook guidelines favor 
self-contained, minimal examples. Adding a signed-receipt composition layer 
would be a separate cookbook entry rather than an expansion of this one.

If you'd like to author a "Governance Evidence and Audit Receipts" cookbook 
showing the receipt layer composing with any gate (including TealTiger), that 
could be a good standalone PR. The separation you're describing would be a 
natural follow-up once this one lands.

nagasatish007:
Still relevant and I'd like to land this — keeping it open. I'll rebase on latest `main` and resolve the current conflict shortly.

To recap for anyone new to the thread: this cookbook fills a gap the existing governance cookbooks don't cover — **input governance at the tool-call boundary**, i.e. deterministic policy evaluation between "model decides" and "tool executes" (<1ms, no second LLM call). "Building Governed AI Agents" covers the Agents SDK + Guardrails and "How to implement LLM guardrails" covers output guardrails; this one is the pre-execution gate.

Per the earlier discussion with @arian-gogani, I've deliberately scoped this to the gating pattern — the signed-receipt / evidence layer is a natural standalone follow-up cookbook rather than an expansion of this one.

Happy to take maintainer feedback on structure or fit. Commenting to reset the stale timer so it doesn't auto-close while I rebase.

PR #2796, opened 2026-06-17
[codex] Add local Codex OTel observability cookbook

henzelmann-oai:
## Summary

- Add a local Codex OpenTelemetry observability cookbook guide under `examples/codex/codex-opentelemetry-observability/`.
- Include a Docker Compose Grafana LGTM stack, OpenTelemetry Collector privacy filters, Grafana dashboard provisioning, a local Codex OTel profile example, and local validation scripts.
- Add the Grafana overview image plus `registry.yaml` and `authors.yaml` metadata so the guide can publish on cookbook.openai.com.

## Why

This gives users a runnable local path for observing Codex logs, traces, and metrics with OpenTelemetry while keeping telemetry opt-in and privacy filtered by default.

## Validation

- `./scripts/validate.sh`

Validation passed for shell syntax, Python tests, source asset validation, registry schema, publication metadata, and Docker Compose configuration. Repository notebook validation was skipped because `nbformat` is not installed in the current local Python environment.


Necmttn:
A useful local-first example would be correlating Codex OTLP output back to the local session transcript.

Show the join keys explicitly: conversation/thread id, event sequence or timestamp, model, token buckets, and whether the signal arrived as OTLP logs, metrics, or traces. That helps users tell "collector is working" from "I can explain which Codex thread spent these tokens".

Generated with ax - https://github.com/Necmttn/ax

PR #2812, opened 2026-06-21
feat(examples): error handling and retry patterns for the OpenAI API

RudrenduPaul:
## Summary

Adds `examples/error_handling_and_retry.ipynb` — a production-focused guide to handling API errors and rate limits gracefully when calling OpenAI.

**The problem**: Simple `try/except` or immediate retries either fail silently, hammer a saturated endpoint, or mask the root cause. Each pattern in this notebook targets a distinct failure mode.

**What the notebook covers:**

| Pattern | Failure mode targeted |
|---|---|
| Error taxonomy table | Knowing which errors to retry vs. fail fast |
| Tenacity exponential backoff + jitter | Transient 429s and 5xx errors |
| Rate-limit header inspection | Capacity monitoring per request |
| Proactive `RateLimitTracker` | Avoid hitting the limit entirely |
| `httpx.Timeout` configuration | Slow responses and connection hangs |
| Async per-task retry isolation | One failed request doesn't cancel others |
| `FallbackClient` (gpt-4o → gpt-4o-mini) | Sustained capacity overflow |
| Structured error logging | Debugging with `request_id` and headers |

**Key implementation details:**
- `chat_with_retry()`: retries `RateLimitError`, `APIError` (5xx), `APIConnectionError` only — auth and validation errors fail fast
- `RateLimitTracker`: reads `x-ratelimit-remaining-requests` / `x-ratelimit-reset-requests` headers, sleeps when `remaining < low_water_mark`
- `FallbackClient.create()`: catches `RateLimitError` on `gpt-4o`, immediately retries with `gpt-4o-mini`, returns which model was used
- `log_error()`: captures `error_type`, `status_code`, `request_id`, `body.error.code`, and all rate-limit headers into a structured dict

Uses `gpt-4o-mini` as default model. No hardcoded API keys. No Anthropic SDK imports.

## Test plan

- [ ] Run all cells top-to-bottom with a valid `OPENAI_API_KEY`
- [ ] Confirm the tenacity retry demo shows simulated error then success
- [ ] Confirm async concurrent cell completes 3 tasks with per-task retry isolation

Built by Rudrendu Paul, developed with Claude Code


RudrenduPaul:
All four items from the Codex automated review have been addressed in commit 6c4113f:

**P0 — registry.yaml entry**: Added entry with slug `error-handling-and-retry`, description, date, author, and tags (`errors`, `retry`, `rate-limits`, `async`, `tenacity`, `completions`).

**P1 — Narrow retries to transient errors**: `RETRYABLE_ERRORS` was narrowed to `(RateLimitError, InternalServerError, APITimeoutError, APIConnectionError)` — the base `APIError` class is no longer used, so auth/400/403/404 errors fail fast as the taxonomy describes.

**P2 — Rate-limit reset header format**: Mock values updated to use duration strings (`"6s"`, `"1s"`) matching the actual OpenAI API header format. The markdown table label was also updated from "ISO-8601 duration" to "Duration string" for clarity.

**P2 — asyncio.run() in Jupyter**: Replaced with top-level `await run_concurrent_batch(...)` — the kernel's existing event loop is reused correctly.

RudrenduPaul:
@codex review

RudrenduPaul:
Rebased onto latest main (resolved a registry.yaml conflict) and addressed all 3 outstanding findings from the last review:

- **Disable SDK retries when Tenacity owns the loop**: real client now instantiated with `max_retries=0`, with a comment explaining why (stacking 6 tenacity attempts on top of the SDK's own default `max_retries=2` could mean up to 18 real HTTP calls per logical retry cycle).
- **Fail fast on quota-exhausted 429s**: `FallbackClient` now checks `error.code == "insufficient_quota"` and re-raises immediately instead of trying the fallback model, since quota exhaustion is account-wide and switching models won't help. Added a new mock demonstration cell showing the fail-fast path.
- **Read error codes from the top-level body**: `log_error` now reads `error.code` directly (which the SDK already unwraps from the response body) instead of looking for a nested `error.body["error"]["code"]` wrapper that doesn't actually exist by the time the exception reaches user code. Fixed the mock error body shape to match the real unwrapped format.

Executed the full notebook end-to-end after the fix (nbconvert --execute), zero errors, all cells including the new demo produce the expected output.

@codex review

RudrenduPaul:
Addressed all 3 findings from the latest review (commit c5fff60):

- **Honor Retry-After before Tenacity retries**: both retry decorators now use a custom wait function that reads the exception's `Retry-After` header as a minimum wait when present, and adds jitter to the exponential back-off otherwise, so concurrent workers don't retry in lockstep.
- **Preserve partial async batch results**: `run_concurrent_batch` now uses `asyncio.gather(..., return_exceptions=True)`. Added a mixed-success demo showing one question failing (non-retryable `AuthenticationError`) while the other two still return their results in the same batch.
- **Sleep for the reset window when throttling**: `RateLimitTracker` now parses `x-ratelimit-reset-requests` (e.g. `"6s"`, `"1m30s"`) and sleeps for that duration instead of a fixed 2s guess. Updated the mock demo to use varying reset durations and confirmed the log output shows 6.0s and 90.0s respectively instead of a constant 2s.

Executed the full notebook end-to-end via `nbconvert --execute` after the fix — zero errors, all new demo cells produce the expected output.

@codex review

RudrenduPaul:
Addressed all 3 findings from the latest review:

- **Track token-limit headroom before claiming throttling**: `RateLimitTracker` now also reads `x-ratelimit-remaining-tokens`/`x-ratelimit-reset-tokens` and sleeps proactively when either requests or tokens run low, since they're independent limit dimensions. Added a demo showing a token-heavy call (healthy request count, low token count) still triggering the proactive sleep.
- **Stop calling reset headers ISO-8601**: fixed both the markdown table and the code comment — these are plain duration strings (`"6s"`, `"1m30s"`), not ISO-8601 (which would look like `"PT6S"`).
- **Refresh stale rate-limit demo output**: re-executed the full notebook so the saved output actually reflects the dynamic sleep durations (6.0s / 90.0s / 12.0s) instead of a leftover fixed `"2s"` from before the earlier fix.

@codex review

RudrenduPaul:
Addressed both findings:

- **Timeout component count**: the prose said `httpx.Timeout` takes three components but the following bullets list four (connect, read, write, pool) — fixed the count.
- **Clear failed setup-cell output**: re-executed the notebook in a clean, isolated virtualenv so the `%pip install` cell actually succeeds instead of recording an `externally-managed-environment` failure while later cells still showed successful imports.

@codex review

RudrenduPaul:
Fixed. \`insufficient_quota\` is an account-level failure requiring billing/plan action — retrying it can never succeed, it just wastes attempts and delays the caller. Replaced the type-based \`retry_if_exception_type(RETRYABLE_ERRORS)\` with a \`retry_if_exception\` predicate (\`_is_retryable\`) that excludes \`RateLimitError\`s carrying \`code="insufficient_quota"\` while still retrying ordinary rate-limit 429s and the other \`RETRYABLE_ERRORS\` types normally. Applied to both the sync and async retry decorators. Added a demo proving it: \`call_count=1\` confirms tenacity makes zero retry attempts on a hard-quota error. Re-executed the full notebook in a clean virtualenv, zero errors.

@codex review

RudrenduPaul:
Addressed both findings:

- **Treat all hard 429 codes as non-retryable**: expanded \`HARD_QUOTA_CODES\` to cover all 5 codes OpenAI documents as billing/quota failures — \`insufficient_quota\`, \`credit_balance_exhausted\`, \`organization_spend_limit_exceeded\`, \`project_spend_limit_exceeded\`, \`organization_usage_limit_exceeded\`. \`FallbackClient\` now checks membership in the same set instead of its own hardcoded single-code comparison, so the retry decorators and the fallback client agree on what's unrecoverable.
- **Add a quota fail-fast path to the decision tree**: added an explicit fail-fast branch for hard-quota \`RateLimitError\`s ahead of the retry/throttle/fallback options, so a reader following the tree as a checklist doesn't recreate the wasted calls the rest of the notebook teaches them to avoid.

Re-verified both existing demos (retry decorator, FallbackClient) still pass with the set-based check, and re-executed the full notebook in a clean virtualenv — zero errors.

@codex review

RudrenduPaul:
Fixed all 3:

- **Qualify RateLimitError retries in the taxonomy**: the rule of thumb now says "RateLimitError — but only when its code is not in HARD_QUOTA_CODES" instead of unconditionally.
- **Match the selective-retry helper to the code**: the bullet now describes \`retry_if_exception(_is_retryable)\` and explains why \`retry_if_exception_type\` alone couldn't do this (it only matches exception classes, not error codes within a class).
- **Describe error.body as the unwrapped error dict**: corrected to say \`message\`/\`code\` are top-level keys on \`error.body\` (also mirrored as the \`error.code\` attribute), not nested under a second wrapper that doesn't exist.

Re-executed the full notebook in a clean virtualenv, zero errors.

@codex review

RudrenduPaul:
Fixed both:

- **Qualify the RateLimitError table row**: qualified it to match the rule of thumb and decision tree — retryable unless `code` is in `HARD_QUOTA_CODES`.
- **Include project-token limit headers**: \`RateLimitTracker\` now also reads \`x-ratelimit-remaining-project-tokens\`/\`x-ratelimit-reset-project-tokens\` and uses whichever token bucket (org/model-level or project-level) is lower as the binding constraint. While implementing this I found and fixed a real bug: without resetting \`self._remaining_tokens\` at the top of each call, a healthy reading could get silently masked by a stale low value cached from a previous call. Added a demo proving the fix correctly picks the binding value (900, not the healthy 150000).

Re-executed the full notebook in a clean virtualenv, zero errors.

@codex review

RudrenduPaul:
Fixed. \`call_with_header_inspection\` (the notebook's first observability pattern) still dropped the project-token headers even though \`RateLimitTracker\` already accounted for them — added \`x-ratelimit-remaining-project-tokens\`/\`x-ratelimit-reset-project-tokens\` to the header table, the \`rate_info\` dict, and the mock demonstration, so a project bound by its own token limit doesn't get misleading headroom reported here.

Re-executed the full notebook in a clean virtualenv, zero errors.

@codex review

RudrenduPaul:
Fixed all 3:

- **Honor retry-after-ms before backing off**: \`_wait_respecting_retry_after\` now mirrors openai-python's own Retry-After parser — \`retry-after-ms\` first, then numeric \`retry-after\` seconds, then HTTP-date format — instead of only reading the plain numeric header. Verified all 4 cases standalone against the SDK's actual parsing order.
- **Record token rate-limit headers**: \`log_error\` now also captures \`x-ratelimit-remaining/reset-tokens\` and the project-scoped variants, so a 429 caused by TPM or project-token exhaustion isn't logged with only request headroom (which alone doesn't explain it).
- **Sleep for every low token bucket**: \`RateLimitTracker\` now sleeps for the longest reset among every token bucket that's individually below the threshold, not just whichever bucket happens to have the smallest absolute remaining count. Added a demo where both buckets are low with different reset times (3s vs 60s) — confirmed it correctly sleeps 60s, not the shorter 3s the old logic would have picked.

Re-executed the full notebook in a clean virtualenv, zero errors.

@codex review

RudrenduPaul:
Fixed. A past HTTP-date (already elapsed, or clock skew) parses to a negative value, and Python's \`float()\` happily accepts \`"inf"\`/\`"nan"\` as valid input for the numeric \`retry-after\` path — either would reach \`time.sleep()\` and either crash with \`ValueError\` or hang forever. Now only a finite, non-negative result is trusted: negatives clamp to 0, \`inf\`/\`nan\` fall through to exponential backoff instead. Verified all 4 cases standalone (past date, inf, nan, normal positive value) before touching the notebook.

Re-executed the full notebook in a clean virtualenv, zero errors.

@codex review

PR #2824, opened 2026-06-24
Retarget Parallel MCP cookbook for an AI-native audience

minskimm:
## What this changes

Retargets the Parallel Search MCP cookbook for an AI-native / YC audience so the examples land with senior technical builders rather than reading as a generic finance lookup.

### Use Case 1 — Grounded Answers
- **New 1.4 baseline cell:** asks the question with *no tools* so GPT-5.5's knowledge gap is visible, then 1.5 answers the **same** question grounded via `web_search`. Makes the value of retrieval legible in one before/after.
- Swapped the Apple SG&A lookup for a fresher, post-cutoff query: **recent YC X25 (Spring 2025) Series A rounds**.
- Moved the 1.8 multi-step streaming example off Apple onto a **Ramp funding-history trace** (still exercises search + fetch + primary-source citations).

### Use Case 2 — Structured Enrichment
- Enriches a short list of **marquee target accounts** instead of a single record.
- **New section 2.6** fans the enrichment out concurrently into a pandas dataframe (the "now scale it" payoff).
- Tightened the recent-signal window from 90 → 7 days.

### Notes
- **Outputs are cleared pending a re-run with GPT-5.5 access** — opening as a draft so the team can execute the notebook and commit real outputs.
- Open questions for reviewers: exact YC query wording, the Ramp choice for 1.8, and the specific logos in the target-account list — all easy to adjust.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/partners/parallel_search_mcp_enrichment/parallel_search_mcp_enrichment.ipynb
+++ b/examples/partners/parallel_search_mcp_enrichment/parallel_search_mcp_enrichment.ipynb
@@ -51,7 +51,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install --quiet --upgrade openai"
+    "%pip install --quiet --upgrade openai pandas"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


minskimm:
Superseded by https://github.com/openai/openai-cookbook/pull/2831. That PR cleared committed outputs and bundled a content rework, which made it hard to review. The replacement is source-only, preserves all outputs, and is scoped to technical polish. Closing this.

PR #2844, opened 2026-07-09
Add avian magnetoreception quantum co-scientist example

steps-re:
### Summary

A community example that casts the model as a hands-on scientific collaborator on a real physics problem: the radical-pair "quantum compass" thought to underlie avian magnetoreception.

GPT does the actual work across the notebook:
- **Designs** a Trotterized circuit for the 3-qubit radical-pair Hamiltonian, choosing the gate decomposition and singlet preparation/readout.
- **Runs** it on a simulator and compares against pre-recorded runs on real IBM quantum hardware (`ibm_kingston`), sweeping the compass-survival probability vs field angle.
- **Interprets** the sweep, and is told the shot-noise test result so it does not invent a hardware-error story the data cannot support.
- **Judges** whether a quantum computer is even the right tool here. The evidence points to no for this observable, and the notebook is explicit that our runs only verify that to N=10.

### Notes

- Runs end-to-end on a laptop from bundled pre-recorded data (`data/`), so no quantum-hardware account or long queue is needed to reproduce it.
- Uses the Responses API. API key is read from `OPENAI_API_KEY`; no secrets in the notebook.
- Executed outputs are kept so the reasoning is visible on the rendered page.
- Step 3 plots both the polarized circuit (what ran on hardware) and the unpolarized average (what a receptor would actually see). The 2.95x swing belongs to the toy's initial state; the physical figure is 1.38x.
- Step 6 ships its generator (`generate_benchmark.py`). It rebuilds `data/frontier_benchmark.json` and has a `--self-test` mode running four controls. Step 4 is still a pre-recorded hardware aggregate with no generating code, and the notebook says so.
- Step 6's table is generated from the data file rather than typed, so it cannot drift.
- Adds `registry.yaml` and `authors.yaml` entries.

### Motivation

The cookbook has no other quantum, physics, or chemistry example (I checked `registry.yaml`). This adds one that uses the model as a working collaborator on a real research question rather than a chat demo, and it ends on a negative result: the model concludes a quantum computer is the wrong tool for this observable and shows the numbers that say so.

---

## For new content

- [x] I have added a new entry in [registry.yaml](https://github.com/openai/openai-cookbook/blob/main/registry.yaml) (and, optionally, in [authors.yaml](https://github.com/openai/openai-cookbook/blob/main/authors.yaml)) so that my content renders on the cookbook website.
- [x] I have conducted a self-review of my content based on the [contribution guidelines](https://github.com/openai/openai-cookbook/blob/main/CONTRIBUTING.md#rubric):
  - [x] Relevance: This content is related to building with OpenAI technologies and is useful to others.
  - [x] Uniqueness: I have searched for related examples in the OpenAI Cookbook, and verified that my content offers new insights or unique information compared to existing documentation.
  - [x] Spelling and Grammar: I have checked for spelling or grammatical mistakes.
  - [x] Clarity: I have done a final read-through and verified that my submission is well-organized and easy to understand.
  - [x] Correctness: The information I include is correct and all of my code executes successfully.
  - [x] Completeness: I have explained everything fully, including all necessary references and citations.


steps-re:
merged main in, the authors.yaml conflict was just three of us appending in the same spot. notebook still validates with nbformat. happy to trim or split this if it's long for a cookbook example.

steps-re:
refreshed it since the repo moved on a fair bit in two months:

- models are now `gpt-5.6-terra` / `gpt-5.6-sol` to match what the recent examples use, and i re-ran the notebook against them.
- found a real gap while doing that: `draw("mpl")` needs `pylatexenc` and it was not in the requirements, so a clean env died on the first circuit drawing. added it.
- these are reasoning models so reasoning tokens eat `max_output_tokens`. the old 2000 budget silently truncated the circuit-design answer mid-token. raised to 6000 and `ask_model` now raises on an `incomplete` response instead of rendering half an answer.
- filled in the PR checklist, it did not exist when i opened this.

checked on qiskit 2.5.2 / aer 0.17.2 / openai 3.8.0. the sweep is seeded so the numbers the prose quotes are unchanged: hardware vs exact MAE 0.0147, worst angle 0.0286, singlet survival swing 2.95x. `check_notebooks.py` passes.

steps-re:
ran a proper adversarial pass on this before asking anyone to review it, and three claims did not hold up. all three were wrong in the flattering direction, so i have corrected them rather than softened the wording.

**the compass swing was inflated.** the circuit starts the nucleus in `|0>`, fully polarized, where real polarization at body temperature is ~1e-10. that turns a 1.38x swing into 2.95x and makes the curve asymmetric (P(0)=0.457 vs P(180)=0.259), i.e. a polar compass. real radical-pair magnetoreception is an inclination compass, and averaging over the nucleus restores P(0)=P(180)=0.358 exactly. step 3 now computes and plots both curves. slightly embarrassing detail: the model's own step 1 answer said "for a mixed nuclear state, repeat for each component and average" and the notebook printed that and then ignored it.

**the memory wall was misattributed.** step 6 said exact simulation dies at N=11 because the state vector grows. the state vector at N=11 is 128 kB, and 64 MB at N=20. what actually stops the solver is the dense O(dim^2) matrix it exponentiates, 1 GB at N=11. that is my implementation, not a property of the problem, and the wrong version was being handed to the model as a stated fact. rewritten, prompt included.

**the typicality evidence was weaker than claimed.** "spread keeps shrinking" was anchored on N=3, which is the global maximum, the spread rises 11.4x from N=1 before it falls. and it is a self-derived quantity, not an accuracy check. since the exact answer has already converged by N=6, agreement past N=10 proves little; using that plateau as an independent check the real scatter is 0.00086 against a self-reported 0.00031.

also: step 4's ladder resolves 1 of 6 points, so it no longer claims a switch-off threshold or the 15.8x ratio (the denominator was noise, other rows give 9x to 106x). it has binomial error bars and a derived null level now instead of a hardcoded 0.015 line. step 3's hardware residuals turn out to be consistent with pure shot noise once you account for 4096 vs 1024 shots (chi2 13.7 on 10 dof, p=0.19), so the NISQ-error reading is retracted, which is a better result anyway. fixed the step 3 observable equation (trace over the nucleus, not a bare overlap, 10x apart at 120 deg) and a backwards endianness comment.

notebook re-executed end to end, `check_notebooks.py` passes.

steps-re:
step 6 was the thesis of the notebook and it rested on a 16-row json with no generating code, which is not something anyone should have to take on trust. it now ships as `generate_benchmark.py` and the data is rebuilt from it.

writing the generator corrected the argument a second time. the exponential cost is not memory, and it is not the dense operator either. the pair is born singlet with the nuclei unpolarized, so the exact singlet yield averages over all 2^N nuclear basis states. that is 2^N propagations, and that is the thing typicality replaces with R=32. measured at N=10: 176.8s vs 5.2s, a 34x speedup where the cost model predicts 2^10/32 = 32x.

i also changed what the section claims. the useful question is not whether the estimator's central value lands close, it is whether its error bar can be trusted. across N=1 to 10 the deviation from exact is inside the reported error bar at every N: 10 of 10 within 2 sigma, largest 1.5, mean 0.8. past N=10 nothing is checked against exact and the notebook says that in those words rather than implying the extrapolation is verified.

`--self-test` runs four controls: convergence to exact as R grows with a calibrated error bar, invariance of the exact average under nuclear basis reordering, a zero-field null returning exactly zero, and time-grid convergence (drift 1.1e-06 against a ~1e-3 sampling error). all pass.

the step 6 table is now generated from the data file instead of typed into markdown, so it cannot drift from the numbers again. units are dimensionless model units and both the prose and the prompt say so, so the benchmark is not read as an absolute effect size.

still outstanding and flagged in the notebook: step 4 is a pre-recorded hardware aggregate with no generating code. i cannot regenerate it without QPU time, so its claims stay the narrowest in the notebook.

steps-re:
last thing, then i will stop touching this. verified it runs for someone who is not me: fresh clone of the branch, a clean venv with only the packages the notebook lists, every cell executes. `generate_benchmark.py --self-test` passes there too (4/4 controls).

rebased on current main, no conflicts, purely additive apart from the registry and authors entries. ready for review whenever someone has time.

PR #2868, opened 2026-07-20
Unarchive Agents SDK memory cookbooks

kkahadze-oai:
## Summary
- Unarchive Context Engineering - Short-Term Memory Management with Sessions.
- Unarchive Context Engineering for Personalization - State Management with Long-Term Memory Notes.

## Why
Both cookbooks cover current Agents SDK memory patterns and do not reference outdated models or APIs. They were included in earlier bulk archive passes without a cookbook-specific reason.

## Impact
The cookbooks return to the main listings and the archived-content warning is removed from their developer-site pages.

## Validation
- Parsed `registry.yaml` with Ruby YAML.
- Ran `git diff --check`.

### DIFF
--- a/examples/agents_sdk/context_personalization.ipynb
+++ b/examples/agents_sdk/context_personalization.ipynb
@@ -220,7 +220,7 @@
    "id": "18a8c00f",
    "metadata": {},
    "source": [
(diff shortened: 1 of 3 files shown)


kkahadze-oai:
@codex review

PR #2880, opened 2026-07-24
Add cookbook: check citation faithfulness in RAG with a zero-token gate

tonydzi:
When a RAG answer cites its sources, the citations fail in ways that read as authoritative and slip past an LLM judge: **fabricated** (a quote in no document), **frankenquote** (real words, a sentence never written contiguously), and **misattributed** (a real quote from the wrong document).

This cookbook shows a *cheap deterministic detector → expensive judge* pattern:

1. Ask the model, via **Structured Outputs** (Responses API + a Pydantic schema), to return each claim with the `document_id` it relies on and a short **verbatim quote**.
2. Run a **0-token verbatim gate** (pure Python — no model, no key) that checks each quote appears in the cited document. This rejects the three failure modes above and runs offline / in CI.
3. Only for quotes that pass the gate, optionally call a **burden-of-proof judge** — fabrications never reach it, so they cost zero tokens.

The offline section executes end-to-end (verified with `nbclient`):
```
OK   [        found]  GPT-4o supports a 128k context.
FLAG [misattributed]  The large embedding model outputs 3072 dims.
FLAG [    not_found]  GPT-4o outputs 3072-dim embeddings.
FLAG [    not_found]  GPT-4o has a 1M token context.
```

Adds `registry.yaml` and `authors.yaml` entries. The gate is inlined so the notebook has no extra dependency; its standalone, framework-agnostic version (gate + burden-of-proof judge) lives at https://github.com/Palo-Alto-AI-Research-Lab/verbatim-citation-gate .

### DIFF
--- a/authors.yaml
+++ b/authors.yaml
@@ -627,3 +627,8 @@ saptob:
   name: "Saptarshi Banerjee"
   website: "https://www.linkedin.com/in/saptarshi-banerjee-83472679"
   avatar: "https://avatars.githubusercontent.com/saptob"
+
+palo-alto-ai-research-lab:
+  name: "Palo Alto AI Research Lab"
+  website: "https://github.com/Palo-Alto-AI-Research-Lab"
+  avatar: "https://avatars.githubusercontent.com/u/194927794?v=4"
--- a/examples/citation_faithfulness_check.ipynb
+++ b/examples/citation_faithfulness_check.ipynb
@@ -1,0 +1,220 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "b40669c5",
+   "metadata": {},
+   "source": [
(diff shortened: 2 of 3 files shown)


tonydzi:
Both review findings are addressed.

**P0 — registry tag.** Confirmed against `registry.yaml`: three entries use `RAG` and exactly one used lowercase `rag`, which was mine. Changed to `RAG` so the page joins the established tag set.

**P1 — `found` framing.** The finding was correct: the gate section claimed more than the gate proves, and it contradicted the burden-of-proof section further down. Reworded so `found` means the citation has cleared *existence and attribution* and is safe to pass on to support checking, explicitly not safe to surface on its own, with a pointer to the judge for the support question. I also relabelled the printed flag from `OK` to `PASS` in the two example cells, since `OK` carried the same overclaim the wording did.

Nothing else in the notebook changed.

PR #3019, opened 2026-08-23
Fix TypeError in EvalsAPI_Audio_Inputs: access Result.passed as attribute

shoemoney:
## Summary

In the results-summary cell of `examples/evaluation/use-cases/EvalsAPI_Audio_Inputs.ipynb`, `item.results[0]["passed"]` raises `TypeError: 'Result' object is not subscriptable`. This changes the line to `item.results[0].passed`.

One-line change. The notebook JSON was edited in place, so the diff is exactly one line with no metadata churn.

## Motivation

The output-item `results` entries are typed `Result` Pydantic models in the current openai SDK (verified on 3.3.1), so the field must be read as an attribute. The same summary cell in the sibling notebook `EvalsAPI_Image_Inputs.ipynb` already uses attribute access on main; this brings the audio notebook in line with it.

This is existing content, not new or relocated, so no `registry.yaml` or `authors.yaml` changes are needed.


sylvesterkaczmarek (commented):
This matches the SDK object shape and the sibling image-input notebook. `results[0]` is a typed result model, so `.passed` is the correct access here; the old subscript form is the broken path.

Nice small fix.

PR #3058, opened 2026-09-04
Reject non-positive parallel processor limits

### LINKED ISSUE #3012: Parallel API processor can hang with non-positive limits
## Problem

`examples/api_request_parallel_processor.py` accepts non-positive values for `max_requests_per_minute`, `max_tokens_per_minute`, and `max_attempts`.

These values can make the processor fail to terminate:

- `max_requests_per_minute <= 0`: request capacity never reaches 1, so a pending request is never scheduled.
- `max_tokens_per_minute <= 0`: token capacity never reaches a positive request cost, so a pending request can wait forever.
- `max_attempts == 0`: the processor decrements `attempts_left` before the request call, making it `-1`; on failure, `if self.attempts_left:` remains truthy for negative values, so the request is requeued indefinitely.

## Expected behavior

Reject invalid non-positive limits at the processor boundary before opening the request file or entering the scheduling loop.

## Suggested fix

Validate:

- `max_requests_per_minute > 0`
- `max_tokens_per_minute > 0`
- `max_attempts >= 1`

with clear `ValueError` messages, leaving all valid inputs unchanged.

sylvesterkaczmarek:
## Summary

Prevent `examples/api_request_parallel_processor.py` from hanging when configured with non-positive request or token limits, or with zero retry attempts.

The processor now validates its limits at the processing boundary before opening the input file or creating an HTTP session:

- require `max_requests_per_minute > 0`
- require `max_tokens_per_minute > 0`
- require `max_attempts >= 1`

Valid configurations are unchanged.

Fixes #3012.

## Motivation

Non-positive request or token capacities can leave a pending request permanently unable to enter the scheduling path. A zero attempt count is decremented to a negative value before the request is made, and negative integers remain truthy in the retry check, which can cause failed requests to be requeued indefinitely.

Rejecting these configurations immediately gives callers a clear error instead of allowing the processor to stall.

### DIFF
--- a/examples/api_request_parallel_processor.py
+++ b/examples/api_request_parallel_processor.py
@@ -107,6 +107,20 @@ from dataclasses import (
 )  # for storing API inputs, outputs, and metadata
 
 
+def validate_processing_limits(
+    max_requests_per_minute: float,
+    max_tokens_per_minute: float,
+    max_attempts: int,
+) -> None:
+    """Reject limits that can prevent the processor from making progress."""
+    if max_requests_per_minute <= 0:
+        raise ValueError("max_requests_per_minute must be greater than 0")
+    if max_tokens_per_minute <= 0:
+        raise ValueError("max_tokens_per_minute must be greater than 0")
+    if max_attempts < 1:
+        raise ValueError("max_attempts must be at least 1")
+
+
 async def process_api_requests_from_file(
     requests_filepath: str,
     save_filepath: str,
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:
Addressed the sub-unit capacity case. Both request and token rate limits now require at least 1, matching the scheduler thresholds so a positive fractional bucket cannot leave a request pending forever. The example compiles successfully with py_compile.

PR #3083, opened 2026-09-11
Update Agents API examples

alfozan:
## Summary
- Update sandbox connections, cleanup, and DigitalOcean setup.
- Improve document-review calculations and count unique specialists in activity reports.
- Require explicit executor credentials.

## Validation
85 Python checks, 29 TypeScript assertions, all TypeScript type checks, and CI pass. Live checks covered five app workflows with sample data, eight sandbox integrations, and six provider controllers (three via deployed signed HTTP requests).

The latest Document Review run produced valid reports but failed on subagent-history HTTP 500s. Sev Bot required an alternate model. Cloudflare, OCI, OpenAI-origin webhook delivery, and real external app integrations remain unverified.


sylvesterkaczmarek (commented):
`validate_review()` never verifies the printed subtotal even though the updated policy requires comparing it with the sum of line items. Add the printed subtotal to `calculation`, verify `sum(quantity * unit_price) == stated_subtotal`, then verify `stated_subtotal + shipping == calculated_total`; add a regression where the final total matches but the printed subtotal is wrong.

cguo-oai (approved):

PR #3090, opened 2026-09-12
Validate printed invoice subtotals in document review

### LINKED ISSUE #3086: Document review example does not validate the printed invoice subtotal
### Problem

The merged document-review example instructs the specialist to compare the sum of invoice line items with the printed subtotal, but `validate_review()` does not receive or validate that subtotal.

It currently verifies only:

- `sum(quantity * unit_price) + shipping == calculated_total`
- `stated_total - calculated_total == difference`

A report can therefore contain line items whose sum disagrees with the invoice's printed subtotal and still pass validation as long as the final total is internally consistent.

### Suggested fix

Include the printed subtotal in `calculation`, then validate:

1. `sum(quantity * unit_price) == stated_subtotal`
2. `stated_subtotal + shipping == calculated_total`
3. `stated_total - calculated_total == difference`

Add a regression where the final total matches but the printed subtotal is wrong.

This was also noted during review of #3083 before merge.

barry166:
## Summary

- Validate that extracted invoice line items add up to the printed subtotal.
- Require the subtotal in the expense-review report schema and verify subtotal plus shipping against the calculated total.

## Motivation

The document-review example previously checked only the final calculated total. A report could therefore pass even when its line items disagreed with the invoice's printed subtotal, as long as the final total was internally consistent.

Fixes #3086

## Testing

- `python3 -m py_compile examples/agents_api/apps/document_review/agent.py`
- `git diff --check`

The repository's runtime dependencies are not installed in this environment, so the full document-review application was not run.

### DIFF
--- a/examples/agents_api/apps/document_review/agent.py
+++ b/examples/agents_api/apps/document_review/agent.py
@@ -89,15 +89,27 @@ def validate_review(report: dict[str, Any], document: str) -> None:
             for item in line_items
         ]
         shipping = Decimal(str(calculation["shipping"]))
+        stated_subtotal = Decimal(str(calculation["stated_subtotal"]))
         stated = Decimal(str(report["amount"]))
         calculated = Decimal(str(calculation["calculated_total"]))
         difference = Decimal(str(calculation["difference"]))
         if not all(
             value.is_finite()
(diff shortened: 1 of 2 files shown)


jazzypayne1997-hub:
Reread

